PR #8, opened 2018-08-24
Open up token audience for authentication

alexeldeib:


lib/login.ts
```diff
@@ -170,8 +173,8 @@ export async function withUsernamePasswordWithAuthResponse(username: string, pas
     await creds.getToken();
     // The token cache gets propulated for all the tenants as a part of building the tenantList.
     tenantList = await buildTenantList(creds);
-    // We dont need to get the subscriptionList if the tokenAudience is graph as graph clients are tenant based.
-    if (!(options.tokenAudience && options.tokenAudience === TokenAudience.graph)) {
+    // We only need to get the subscriptionList if the tokenAudience is for a management client.
+    if (options.tokenAudience && options.tokenAudience === options.environment.activeDirectoryResourceId) {
```
amarzavery: This check also needs to happen for interactive login. I think we missed adding it over there while doing the initial port from ms-rest-azure.

After building the tenantList, we get subscriptions from different tenants. This exercise populates the token cache with tokens across all the tenants. However, this exercise should only be done for managementPlane token audience (whether logging in via username/password, service/principal or interactive.

So the check happening over [here](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest-azure/lib/login.js#L241-L243) in the old runtime needs to happen in this project over [here](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L464) as well. Basically pass an empty list of tenants to getSubscriptionFromTenants if the tokenAudience is not a managementplane token audience.

amarzavery: ignore this comment. I just saw that you have taken care of this below.

lib/login.ts
```diff
@@ -455,7 +455,7 @@ export async function withInteractiveWithAuthResponse(options?: InteractiveLogin
         if (error) {
           return reject(error);
         }
-        interactiveOptions.username = tokenResponse.userId;
+        interactiveOptions.userName = tokenResponse.userId;
```
alexeldeib: @amarzavery @RikkiGibson FYI, I think this fixes interactive login. ADAL couldn't match the token in the cache because it used user@example.com as cache key, while the token had been acquired correctly with the user's email.

amarzavery: Agreed. Thanks for fixing this. This is absolutely correct.

RikkiGibson: This is the most JavaScript bug ever.

amarzavery:
1. While looking at that source code in the old runtime over [here](https://github.com/Azure/azure-sdk-for-node/blob/8691046c2ca738552d815f6967f5f54b04840276/runtime/ms-rest-azure/lib/login.js#L192), I found that we pass the tokenAudience if provided otherwise `environment.activeDirectoryResourceId` as the `resource` to `acquireTokenWithDeviceCode` and `acquireUserCode`. I did not see that part being ported over.

2. Another [issue](https://github.com/Azure/azure-sdk-for-node/issues/2002) that is tangential to the problem you are solving  was reported  in ms-rest-azure. adal-node would intermittently send an error `autorization_pending` and the work around was to trap that error, wait for sometime and try again till the error goes away. This was fixed over [here](https://github.com/Azure/azure-sdk-for-node/blob/8691046c2ca738552d815f6967f5f54b04840276/runtime/ms-rest-azure/lib/login.js#L196-L210) in ms-rest-azure. Is it possible for you to patch this as well, while you are improving the interactive login auth experience in this library. It would greatly benefit everyone and all issues pertaining to auth will be resolved. If you think, it will take too much time, then that is totally fine as well. Thanks for taking time and sending a PR 👍

RikkiGibson:
THANK YOU @alexeldeib!

typings/lib/credentials/tokenCredentialsBase.d.ts
```diff
@@ -31,7 +30,7 @@ export declare abstract class TokenCredentialsBase {
     tokenCache: any;
     protected readonly isGraphContext: boolean;
     protected readonly authContext: any;
-    constructor(clientId: string, domain: string, tokenAudience?: TokenAudience | undefined, environment?: {
+    constructor(clientId: string, domain: string, tokenAudience?: string | undefined, environment?: {
```
amarzavery: This should still be TokenAudience | undefined. Since over here we have the [TokenAudience](https://github.com/Azure/ms-rest-nodeauth/pull/8/files#diff-b7899d05e11e38cd904871afb5198981R8) type defined.

Moreover, we should keep the code backwards compatible. ms-rest-azure supported `"graph" | "batch" | undefined`. Can you please add support for batch as well?

I think [this change](https://github.com/Azure/azure-sdk-for-node/commit/126ee8b3aebadd00fd904da3478167fb71ee738c#diff-f3e1ce7c2241aa07c078236e8ce3b25d) should also happen in [ms-rest-azure-env](https://github.com/Azure/ms-rest-azure-env)

lib/credentials/tokenCredentialsBase.ts
```diff
@@ -31,7 +31,7 @@ export abstract class TokenCredentialsBase {
       throw new Error("domain must be a non empty string.");
     }
 
-    if (this.tokenAudience === TokenAudience.graph) {
+    if (this.tokenAudience === "graph") {
       this.isGraphContext = true;
```
amarzavery: Since we only supported graph, this property `isGraphContext ` made sense. I believe don't need this anymore.

lib/credentials/tokenCredentialsBase.ts
```diff
@@ -47,10 +47,14 @@ export abstract class TokenCredentialsBase {
...
 
   protected getActiveDirectoryResourceId(): string {
-    const resource = this.isGraphContext
+    let resource = this.isGraphContext
       ? this.environment.activeDirectoryGraphResourceId
       : this.environment.activeDirectoryResourceId;
 
+    if (this.tokenAudience) {
+      resource =  this.tokenAudience;
+    }
```
amarzavery: This would be much simpler now
```ts
protected getActiveDirectoryResourceId(): string {
  let resource = this.environment.activeDirectoryResourceId;
  if (this.tokenAudience) {
    resource = this.tokenAudience;
    if (this.tokenAudience === "graph") {
      resource = this.environment.activeDirectoryGraphResourceId;
    } else if (this.tokenAudience === "graph") {
     resource = this.environment.batchResourceId;
    }
  }

  return resource;
}
```

typings/lib/credentials/tokenCredentialsBase.d.ts
alexeldeib: The compiler is being smart and simplifying the union type of string with string automatically to remove the enum. I fiddled around with ways to trick the compiler (union type of string with enum of string, etc) but didn't get much.

I can manually fix the typings to use TokenAudience, it will cause drift in the future for generation. Just a thought for reference. 

I'll add batch too.

amarzavery:
I see. Then just having graph | batch | string | undefined would work

alexeldeib:
Actually, I think something weirder is happening. The compiler uses TokenAudience on the [non-abstract implementations](https://github.com/Azure/ms-rest-nodeauth/blob/3618e5191300f948fee851522bc6a3075ae37403/typings/lib/credentials/userTokenCredentials.d.ts#L22) of TokenCredentialsBase, but it doesn't want to use accept TokenAudience type on the abstract base class?

Might be something weird on my machine/locally; happy to leave it whichever way you prefer, since it doesn't seem like there's an inherent issue.

lib/credentials/applicationTokenCredentials.ts
```diff
@@ -16,8 +16,8 @@ export class ApplicationTokenCredentials extends TokenCredentialsBase {
    * @param {string} clientId The active directory application client id.
    * @param {string} domain The domain or tenant id containing this application.
    * @param {string} secret The authentication secret for the application.
-   * @param {string} [tokenAudience] The audience for which the token is requested. Valid value is "graph". If tokenAudience is provided
-   * then domain should also be provided its value should not be the default "common" tenant. It must be a string (preferrably in a guid format).
+   * @param {string} [tokenAudience] The audience for which the token is requested. Valid values are 'graph' or any other resource like 'https://vault.azure.com/'.
```
amarzavery: graph or batch or ...

lib/credentials/deviceTokenCredentials.ts
```diff
@@ -20,8 +20,8 @@ export class DeviceTokenCredentials extends TokenCredentialsBase {
    * @param {string} [clientId] The active directory application client id.
    * @param {string} [domain] The domain or tenant id containing this application. Default value is "common"
    * @param {string} [username] The user name for account in the form: "user@example.com".
-   * @param {string} [tokenAudience] The audience for which the token is requested. Valid value is "graph". If tokenAudience is provided
-   * then domain should also be provided and its value should not be the default "common" tenant. It must be a string (preferrably in a guid format).
+   * @param {string} [tokenAudience] The audience for which the token is requested. Valid values are 'graph' or any other resource like 'https://vault.azure.com/'.
+   * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferrably in a guid format).
```
amarzavery: same as above.

lib/credentials/tokenCredentialsBase.ts
```diff
@@ -31,26 +30,25 @@ export abstract class TokenCredentialsBase {
...
-    const resource = this.isGraphContext
-      ? this.environment.activeDirectoryGraphResourceId
-      : this.environment.activeDirectoryResourceId;
-
+    let resource = this.environment.activeDirectoryResourceId;
+    if (this.tokenAudience) {
+      resource = this.tokenAudience;
+      if (this.tokenAudience === "graph") {
+        resource = this.environment.activeDirectoryGraphResourceId;
+      } else if (this.tokenAudience === "graph") {
```
amarzavery: === "batch"

amarzavery: I think I messed it up while providing the code snippet

alexeldeib: I fixed it, and then re copied it wrong 😄

lib/login.ts
```diff
@@ -421,10 +430,17 @@ export async function withInteractiveWithAuthResponse(options?: InteractiveLogin
   interactiveOptions.context = authContext;
   let userCodeResponse: any;
   let creds: DeviceTokenCredentials;
-  const getUserCode = new Promise<any>((resolve, reject) => {
-    return authContext.acquireUserCode(interactiveOptions.environment.activeDirectoryResourceId, interactiveOptions.clientId, interactiveOptions.language, (err: Error, userCodeRes: any) => {
+
+  function tryAcquireToken(interactiveOptions: InteractiveLoginOptions, resolve: any, reject: any) {
```
alexeldeib: @amarzavery This is about as close as I could get to the other implementation, but I haven't hit the mentioned issue or have a repro, so PSA that this is untested from me.

amarzavery: Cool. No worries. Looking at the changes it looks good to me.

amarzavery:
Referencing this PR https://github.com/Azure/ms-rest-azure-env/pull/2 as it is related to this change.
@alexeldeib - Please update the min version for ms-rest-azure-env dependency to "^0.1.1"

alexeldeib:
@amarzavery thanks in return for the quick turnaround! Glad I could contribute 😄

PR #18, opened 2018-09-19
Add support for KeyVault credentials

kpajdzik:


lib/credentials/keyVaultCredentials.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  signRequest(webResource: WebResource): Promise<WebResource> {
+    // authentication is provided by the createSigningFilter method.
+    return Promise.resolve(undefined);
+  }
+
+  createSigningFilter(): (resource: WebResource, next: Function, callback: ServiceCallback<any>) => any {
+    const self = this;
```
daschult: Is this needed for TypeScript? I thought `this` could be used without creating `self` variables.

lib/credentials/keyVaultCredentials.ts
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    if (!this.authenticator) {
+      this.authenticator = this._authenticatorMapper(this.credentials);
+    }
+  }
+
+  signRequest(webResource: WebResource): Promise<WebResource> {
+    // authentication is provided by the createSigningFilter method.
+    return Promise.resolve(undefined);
```
RikkiGibson: Does it make sense to implement ServiceClientCredentials if this is just a dummy method? It doesn't make sense to ever call this `signRequest`, does it?

lib/credentials/keyVaultCredentials.ts
kpajdzik: Yes, but we need to use => notation, right? I didn't want to change too much in the code.

RikkiGibson: Riight, once you're in the `function() {}` scope, `this` isn't from the KeyVaultCredentials any more.

lib/credentials/keyVaultCredentials.ts
kpajdzik: That's true. It's mirrored from Node SDK and I think it's worth keeping the object structure. What do you think?

kpajdzik: I've created an issue for that: https://github.com/Azure/ms-rest-nodeauth/issues/20

PR #22, opened 2018-09-27
Add KeyVaultFactory

kpajdzik:


lib/credentials/keyVaultFactory.ts
```diff
@@ -0,0 +1,81 @@
...
+import { DeviceTokenCredentials } from "./deviceTokenCredentials";
+import { MSIAppServiceTokenCredentials } from "./msiAppServiceTokenCredentials";
+import { MSITokenCredentials } from "./msiTokenCredentials";
+import { MSIVmTokenCredentials } from "./msiVmTokenCredentials";
+import { TokenCredentialsBase } from "./tokenCredentialsBase";
+import { UserTokenCredentials } from "./userTokenCredentials";
+import { AuthenticationContext, TokenResponse, ErrorResponse } from "adal-node";
+import { Authenticator } from "ms-rest-js";
+
+export class KeyVaultFactory {
```
daschult: TypeScript generally doesn't use classes when they only contain static functions. I'd recommend removing this class and just exporting the public static functions from this file.

lib/credentials/keyVaultFactory.ts
```diff
@@ -0,0 +1,79 @@
...
+function _convert(credentials: MSITokenCredentials): MSITokenCredentials {
+  if (credentials instanceof MSIAppServiceTokenCredentials) {
+    return new MSIAppServiceTokenCredentials({
+      msiEndpoint: credentials.msiEndpoint,
+      msiSecret: credentials.msiSecret,
+      msiApiVersion: credentials.msiApiVersion,
+      resource: credentials.resource
+    });
+  } else if (credentials instanceof MSIVmTokenCredentials) {
+    return new MSIVmTokenCredentials({
```
RikkiGibson: What does this do? Has this code just pretty much always been here?

kpajdzik: Yes, I tried to optimize it but I'm a little scared without unit tests.

RikkiGibson: Yeah, it's hard to test authentication code with anything besides ad-hoc testing.

PR #23, opened 2018-10-02
Update Authenticator implementation in KeyVaultFactory

kpajdzik:


lib/credentials/keyVaultFactory.ts
```diff
@@ -39,21 +39,21 @@ function _convert(credentials: MSITokenCredentials): MSITokenCredentials {
...
       if (tokenResponse.error) {
-        return callback(tokenResponse.error, undefined);
+        return reject(tokenResponse.error);
       }
 
       tokenResponse = tokenResponse as TokenResponse;
       // Calculate the value to be set in the request's Authorization header and resume the call.
       const authorizationValue = tokenResponse.tokenType + " " + tokenResponse.accessToken;
-      return callback(undefined, authorizationValue);
+      resolve(authorizationValue);
```
daschult: `return resolve(authorizationValue);`?

lib/credentials/keyVaultFactory.ts
```diff
@@ -39,21 +39,21 @@ function _convert(credentials: MSITokenCredentials): MSITokenCredentials {
 }
 
 function _createAuthenticatorMapper(credentials: MSITokenCredentials): Authenticator {
```
daschult: Do you need to update the type signature for `Authenticator` because of this work? It seems like it previously was a function that took a callback, but now its a function that returns a Promise.

kpajdzik: I've changed it as a part of request policy work and now I'm adjusting this code.

PR #36, opened 2018-11-14
Add rollup

### LINKED ISSUE #35: Add rollup-plugin-sourcemaps

ghost:
Resolves https://github.com/Azure/ms-rest-nodeauth/issues/35.


kpajdzik (approved):

sergey-shandar (approved):

amarzavery:
This library is not intended for use in the browser. I assume rollup is used to produce es6 modules. Is that correct?

daschult:
Rollup is just used as a bundler to reduce the size of your distributed package. That is very useful in the browser scenario, but it could also be nice in the Node.js world. We're already using rollup in ms-rest-js for both the browser and Node.js scenarios, so this PR is really about bringing parity to this repository from ms-rest-js.

amarzavery:
I see. That sounds cool

Issue #37, opened 2018-11-17
Cannot find module '@azure/ms-rest-nodeauth'.


TechRazor:
When testing out the new sdk's I get the following error on the @azure/ms-rest-nodeauth module as the package on npm is missing the dist folder.

```typescript
filename.ts:1:33 - error TS2307: Cannot find module '@azure/ms-rest-nodeauth'.
```

Temporary workaround I'm using:

```bash
cd <my-project-directory>
cd './node_modules/\@azure/ms-rest-nodeauth'
npm run build
```

then in my .ts file use this for the import:

```typescript
import * as msRestNodeAuth from './node_modules/@azure/ms-rest-nodeauth/dist/msRestNodeAuth';
```

kpajdzik:
This is fixed in [version 0.8.2](https://www.npmjs.com/package/@azure/ms-rest-nodeauth/v/0.8.2). Apologies for the problem!

TechRazor:
@kpajdzik thanks for the quick fix!

PR #39, opened 2018-11-19
Fix incorrect path to the main file

kpajdzik:


kpajdzik:
Fixes https://github.com/Azure/ms-rest-nodeauth/issues/37

Issue #49, opened 2019-03-14
Change login.withServicePrincipalSecret()'s return type to ApplicationTokenCredentials


daschult:
**Is your feature request related to a problem? Please describe.**
Previously [login.withServicePrincipalSecret()](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L616) returned an [ApplicationTokenCredentials](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/credentials/applicationTokenCredentials.ts#L9) object. After some refactoring took place, the return type changed to [TokenCredentialsBase](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/credentials/tokenCredentialsBase.ts#L10), even though the return value is always an [ApplicationTokenCredentials](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/credentials/applicationTokenCredentials.ts#L9).

**Describe the solution you'd like**
It would be good if [login.withServicePrincipalSecret()](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L616) could go back to returning an [ApplicationTokenCredentials](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/credentials/applicationTokenCredentials.ts#L9). The easiest way to do this would be to cast the return value as an [ApplicationTokenCredentials](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/credentials/applicationTokenCredentials.ts#L9) object.

**Describe alternatives you've considered**
You could also change [AuthResponse](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L90) to be a generic type where the type parameter is the type of the [credentials](https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L94) property. That way all of the `login...()` methods could get accurate type information about their return value.

bterlson:
@daschult is this a fix we can make relatively soon? We're looking to migrate Event Hub libraries to the shiny new auth libs and this change will impact existing customers...

ramya-rao-a:
FYI, I have submitted a PR to fix this issue. Please see https://github.com/Azure/ms-rest-nodeauth/pull/50

ramya-rao-a:
This issue can be closed now as the PR #50 has been merged and version 0.9.2 of this library has the fix

PR #50, opened 2019-03-26
Use the right return types when returning promise

ramya-rao-a:
This PR attempts to fix the issue reported in #49 

The return types for calls using interactive login, user name/ password and service principal return the right types of credentials when the callback pattern is used.
But, when the promise pattern is used, the base class `TokenCredentialsBase` is being returned which causes type issues.

This PR fixes this by updating the return type in the promise patterns to return the right type

cc @bterlson

### DIFF
--- a/lib/login.ts
+++ b/lib/login.ts
@@ -563,8 +563,8 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
  *             @resolve {DeviceTokenCredentials} The DeviceTokenCredentials object.
  *             @reject {Error} - The error object.
  */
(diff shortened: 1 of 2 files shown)


kpajdzik:
LGTM. Let's wait for Dan's review.

ramya-rao-a:
Thanks @kpajdzik and @daschult 
What would be the next steps here to get this PR merged and have a new version released?

daschult:
This change has been published in https://www.npmjs.com/package/@azure/ms-rest-nodeauth/v/0.9.2.

ramya-rao-a:
Thanks @daschult!

PR #52, opened 2019-04-04
Update dependencies fixes #51

### LINKED ISSUE #51: Cannot use storage account features because of outdated @azure/ms-rest-js version
**Package Version**: 0.9.2
- [x] **nodejs**
   - **nodejs version**: 10.15.0
   - **os name/version**: OSX 10.14.3
- [ ] **browser**
    - **name/version**: ____
**Describe the bug**
Error serializing payload when createing blob containers on storage accounts using @azure/arm-storage

Underlying issue in @azure/ms-rest-js is fixed, but @azure/ms-rest-nodeauth is using an outdated version of@azure/ms-rest-js. Version should be updated to 1.8.1:
https://github.com/Azure/ms-rest-nodeauth/blob/master/package.json#L42

**To Reproduce**
See https://github.com/Azure/azure-sdk-for-js/issues/1588

**Expected behavior**
Can create storage account containers when authenticating using @azure/ms-rest-nodeauth

**Screenshots**
n/a

**Additional context**
n/a

amclin:
Updates the dependencies for latest @azure/ms-rest-js

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   ],
   "dependencies": {
     "@azure/ms-rest-azure-env": "^1.1.0",
-    "@azure/ms-rest-js": "^1.1.0",
+    "@azure/ms-rest-js": "^1.8.1",
     "adal-node": "^0.1.22"
   },
   "license": "MIT",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "email": "azsdkteam@microsoft.com",
     "url": "https://github.com/Azure/ms-rest-nodeauth"
   },
-  "version": "0.9.2",
+  "version": "0.9.3",
   "description": "Azure Authentication library in node.js with type definitions.",
   "tags": [
     "node",


package.json
```diff
@@ -39,7 +39,7 @@
   ],
```
daschult: Please update the version in this file to 0.9.3.

package.json
```diff
@@ -39,7 +39,7 @@
...
   "dependencies": {
     "@azure/ms-rest-azure-env": "^1.1.0",
-    "@azure/ms-rest-js": "^1.1.0",
+    "@azure/ms-rest-js": "^1.8.1",
```
amarzavery: Well this is a good change. But even without this change, things should be working fine. We are using the **^** sign which tells npm to install the highest available version in the Major.x.x range.

Isn't that what you see? What is the version of node.js and npm you are using?
At times package-lock.json can create issues by directing npm to install old versions. Deleting package-lock.json file and the node_modules folder in your project should help solve the problem.

Accepting this change would be better anyways.

package.json
amclin: Is that not handled by your CI pipeline? I wouldn't want to presume what your versioning rules are.

package.json
amclin: NPM doesn't do that for dependencies of packages because package-lock.json is not enforced when doing `npm install` on a clean workspace. Plus, `npm install` overwrites package.json, blowing away any settings that may have been manually specified there.

I'm using Node 10.15.0 with NPM 6.4.1 with fresh workspaces.

package.json
daschult: We update our version numbers by hand. Please update it to 0.9.3.

daschult: Thanks!

PR #53, opened 2019-05-05
Implemented ServicePrincipal login with certificate

amarzavery:
- Fixes https://github.com/Azure/ms-rest-azure-js/issues/37, https://github.com/Azure/azure-sdk-for-node/issues/3101

### DIFF
--- a/lib/credentials/applicationTokenCertificateCredentials.ts
+++ b/lib/credentials/applicationTokenCertificateCredentials.ts
@@ -6,7 +6,7 @@ import { createHash } from "crypto";
 import { ApplicationTokenCredentialsBase } from "./applicationTokenCredentialsBase";
 import { Environment } from "@azure/ms-rest-azure-env";
 import { AuthConstants, TokenAudience } from "../util/authConstants";
-import { TokenResponse, ErrorResponse } from "adal-node";
+import { TokenResponse, ErrorResponse, TokenCache } from "adal-node";
 import { AzureTokenCredentialsOptions } from "../login";
 
 export class ApplicationTokenCertificateCredentials extends ApplicationTokenCredentialsBase {
@@ -34,15 +34,15 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
     thumbprint: string,
     tokenAudience?: TokenAudience,
     environment?: Environment,
-    tokenCache?: any
+    tokenCache?: TokenCache
   ) {
-    if (!Boolean(certificate) || typeof certificate.valueOf() !== "string") {
+    if (!certificate || typeof certificate.valueOf() !== "string") {
       throw new Error("certificate must be a non empty string.");
     }
-    if (!Boolean(thumbprint) || typeof thumbprint.valueOf() !== "string") {
+    if (!thumbprint || typeof thumbprint.valueOf() !== "string") {
       throw new Error("thumbprint must be a non empty string.");
     }
-    super(clientId, domain, tokenAudience, environment as any, tokenCache);
+    super(clientId, domain, tokenAudience, environment, tokenCache);
 
     this.certificate = certificate;
     this.thumbprint = thumbprint;
@@ -60,8 +60,8 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
       if (error.message.startsWith(AuthConstants.SDK_INTERNAL_ERROR)) {
         return Promise.reject(error);
       }
-      const resource = this.getActiveDirectoryResourceId();
       return new Promise((resolve, reject) => {
+        const resource = this.getActiveDirectoryResourceId();
         this.authContext.acquireTokenWithClientCertificate(
           resource,
           this.clientId,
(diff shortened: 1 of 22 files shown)


lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+  public constructor(
+    clientId: string,
+    domain: string,
+    certificate: string,
+    thumbprint: string,
+    tokenAudience?: TokenAudience,
+    environment?: Environment,
+    tokenCache?: any
+  ) {
+    if (!Boolean(certificate) || typeof certificate.valueOf() !== "string") {
```
kpajdzik: Why cast to `Boolean`?

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+      throw new Error("certificate must be a non empty string.");
+    }
+    if (!Boolean(thumbprint) || typeof thumbprint.valueOf() !== "string") {
```
kpajdzik: Same here

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+      throw new Error("thumbprint must be a non empty string.");
+    }
+    super(clientId, domain, tokenAudience, environment as any, tokenCache);
```
kpajdzik: What's the parents class type here?

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  public async getToken(): Promise<TokenResponse> {
+    try {
+      const tokenResponse = await this.getTokenFromCache();
+      return tokenResponse;
+    } catch (error) {
+      if (error.message.startsWith(AuthConstants.SDK_INTERNAL_ERROR)) {
+        return Promise.reject(error);
+      }
+      const resource = this.getActiveDirectoryResourceId();
```
kpajdzik: Can we put this line inside of the promise?

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+      !Boolean(certificateStringOrFilePath) ||
+      typeof certificateStringOrFilePath.valueOf() !== "string"
+    ) {
+      throw new Error(
+        "'certificateStringOrFilePath' must be a non empty string."
+      );
+    }
+    let certificate: string = certificateStringOrFilePath;
+    if (!certificateStringOrFilePath.startsWith("-----BEGIN")) {
+      certificate = readFileSync(certificateStringOrFilePath, "utf8");
```
kpajdzik: Are we fine with sync version here?

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+  ): ApplicationTokenCertificateCredentials {
+    if (
+      !Boolean(certificateStringOrFilePath) ||
+      typeof certificateStringOrFilePath.valueOf() !== "string"
+    ) {
+      throw new Error(
+        "'certificateStringOrFilePath' must be a non empty string."
+      );
+    }
+    let certificate: string = certificateStringOrFilePath;
```
kpajdzik:
```suggestion
    const certificate: string = certificateStringOrFilePath.startsWith("-----BEGIN") ? certificateStringOrFilePath : readFileSync;
```
? We can use const in this case.

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -0,0 +1,137 @@
...
+    if (!certificateStringOrFilePath.startsWith("-----BEGIN")) {
+      certificate = readFileSync(certificateStringOrFilePath, "utf8");
+    }
+    const CERT_PATT = /(-+BEGIN CERTIFICATE-+)(\n\r?|\r\n?)([A-Za-z0-9\+\/\n\r]+\=*)(\n\r?|\r\n?)(-+END CERTIFICATE-+)/;
```
kpajdzik: Why this casing? What's PATT?

daschult: PATTern

lib/credentials/applicationTokenCredentialsBase.ts
```diff
@@ -0,0 +1,96 @@
...
+      // Remove the stale token from the tokencache. ADAL gives the same error message "Entry not found in cache."
+      // for entry not being present in the cache and for accessToken being expired in the cache. We do not want the token cache
+      // to contain the expired token, we clean it up here.
+      const status = await self.removeInvalidItemsFromCache({
+        _clientId: self.clientId
+      });
+      if (status.result) {
+        return Promise.reject(error);
+      }
+      const msg =
```
kpajdzik:
```suggestion
      const message =
```
(sorry ;))

lib/credentials/applicationTokenCredentialsBase.ts
```diff
@@ -0,0 +1,96 @@
...
+        status && status.details && status.details.message
+          ? status.details.message
+          : status.details;
+      return Promise.reject(
+        new Error(
+          AuthConstants.SDK_INTERNAL_ERROR +
+            " : " +
+            "critical failure while removing expired token for service principal from token cache. " +
+            msg
```
kpajdzik:
```suggestion
            message
```

kpajdzik:
Sorry for all the comments not related to your changes. Should we think about writing some tests about it?

lib/credentials/applicationTokenCertificateCredentials.ts
amarzavery: ApplicationTokenCredentialsBase

lib/credentials/applicationTokenCertificateCredentials.ts
amarzavery: Done

lib/credentials/applicationTokenCertificateCredentials.ts
amarzavery: Yup. It's a small file. readFileSync should be super quick.

lib/credentials/applicationTokenCertificateCredentials.ts
amarzavery: We won't be able to change the value if the `if` condition is true in the following line.

lib/credentials/applicationTokenCertificateCredentials.ts
kpajdzik: I meant why do we need to cast as any?

lib/credentials/applicationTokenCertificateCredentials.ts
kpajdzik: We don't need that line in this case, right?

amarzavery: Oh yes! We can do something like this:

```js
if (!certificateStringOrFilePath.startsWith("-----BEGIN")) {
      certificateStringOrFilePath = readFileSync(
        certificateStringOrFilePath,
        "utf8"
      );
    }
    const certificatePattern = /(-+BEGIN CERTIFICATE-+)(\n\r?|\r\n?)([A-Za-z0-9\+\/\n\r]+\=*)(\n\r?|\r\n?)(-+END CERTIFICATE-+)/;
    const matchCert = certificateStringOrFilePath.match(certificatePattern);
```

lib/credentials/applicationTokenCertificateCredentials.ts
amarzavery: Ah I see. That makes sense. Copied this part from what was there previously. Now that we have  the type for Environment, will use that. Will also use the typedefinition for TokenCache, since we now have type definitions for adal-node.

PR #56, opened 2019-05-16
 minor updates and review changes

amarzavery:


lib/credentials/azureCliCredentials.ts
```diff
@@ -117,6 +111,21 @@ export class AzureCliCredentials implements TokenClientCredentials {
     return Promise.resolve(webResource);
   }
 
+  private _hasTokenExpired(): boolean {
```
kpajdzik: Do we still follow `_` pattern in other parts of the code?

amarzavery: I still follow this pattern. Not sure if we are consistent with this pattern.

daschult: No, we don't use this anywhere else that I'm aware of. The underscore prefix is reserved for unused parameters.

PR #61, opened 2019-05-22
Add TokenCredential and BearerTokenAuthenticationPolicy

daviwil:
This PR kicks off the new Azure.Identity TypeScript library implementation by temporarily introducing the `TokenCredential` abstract class and `BearerTokenAuthenticationPolicy` class which can apply a token to requests in `ms-rest-js`'s HTTP request pipeline.

## Questions for Reviewers

- Is this the right library/folder to add these classes?  The policy implementation *seems* like it belongs with the other policies in [`ms-rest-js/lib/policies`](https://github.com/Azure/ms-rest-js/tree/master/lib/policies).  Perhaps this low-level implementation belongs in that library instead since it isn't specific to Node.js or the browser?
- Are the comments sufficient?  Are there any posted guidelines for how JSDoc comments should be written?
- The Prettier extension in VS Code has been auto-formatting these files when I save them.  Does the formatting look generally correct?

## Tasks

- [x] Add tests for the policy implementation

/cc @schaabs @bterlson

### DIFF
--- a/lib/identity/bearerTokenAuthenticationPolicy.ts
+++ b/lib/identity/bearerTokenAuthenticationPolicy.ts
@@ -1,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { TokenCredential } from "./tokenCredential";
+import {
+  BaseRequestPolicy,
+  RequestPolicy,
+  RequestPolicyOptions,
+  Constants,
+  HttpHeaders,
+  HttpOperationResponse,
+  WebResource
+} from "@azure/ms-rest-js";
+
+/**
+ *
+ * Provides a RequestPolicy that can request a token from a TokenCredential
+ * implementation and then apply it to the Authorization header of a request
+ * as a Bearer token.
+ *
+ */
+export class BearerTokenAuthenticationPolicy extends BaseRequestPolicy {
+  /**
+   * Creates a new BearerTokenAuthenticationPolicy object.
+   *
+   * @param nextPolicy The next RequestPolicy in the request pipeline.
+   * @param options Options for this RequestPolicy.
+   * @param credential The TokenCredential implementation that can supply the bearer token.
+   * @param scope The scope for which the bearer token applies.
+   */
+  constructor(
+    nextPolicy: RequestPolicy,
+    options: RequestPolicyOptions,
+    private credential: TokenCredential,
+    private scope: string
+  ) {
+    super(nextPolicy, options);
+  }
+
+  /**
+   * Applies the Bearer token to the request through the Authorization header.
+   * @param webResource
+   */
+  public async sendRequest(
+    webResource: WebResource
+  ): Promise<HttpOperationResponse> {
+    if (!webResource.headers) webResource.headers = new HttpHeaders();
+    const token = await this.credential.getToken(
+      [this.scope],
+      webResource.abortSignal
+    );
+    webResource.headers.set(
+      Constants.HeaderConstants.AUTHORIZATION,
+      `Bearer ${token}`
+    );
+    return this._nextPolicy.sendRequest(webResource);
+  }
+}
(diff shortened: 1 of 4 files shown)


lib/identity/tokenCredential.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { AbortSignalLike } from "@azure/ms-rest-js";
+
+/**
+ * Represents a credential capable of providing an authentication token.
+ */
+export abstract class TokenCredential {
```
amarzavery: We already have an interface named TokenCredential. Having an abstract class with the same name will be super confusing.

lib/identity/bearerTokenAuthenticationPolicy.ts
```diff
@@ -0,0 +1,58 @@
...
+} from "@azure/ms-rest-js";
+
+/**
+ *
+ * Provides a RequestPolicy that can request a token from a TokenCredential
+ * implementation and then apply it to the Authorization header of a request
+ * as a Bearer token.
+ *
+ */
+export class BearerTokenAuthenticationPolicy extends BaseRequestPolicy {
```
amarzavery: The policy should probably be in `@azure/ms-rest-js`

kpajdzik:
Please update version in package.json after you address feedback from @amarzavery so the new version of package will get published.

lib/identity/tokenCredential.ts
daviwil: I agree, it could be confusing.  The plan right now based on a conversation with @bterlson and @schaabs is to add these to classes to an existing library so that we can start using them in client libs and then eventually be move them over to the new Azure.Identity library once it gets created.

lib/identity/bearerTokenAuthenticationPolicy.ts
daviwil: @schaabs Do you think I should move both classes over to that `ms-rest-js`?  They're both high-level enough that they don't need to be in the Node-specific library anyway.  Having the new policy class included with the others in that library makes a lot of sense.

daviwil:
@kpajdzik If I export these classes into the public interface of the library, would that entail a minor version bump?  I assume this library follows SemVer 2.0.

kpajdzik:
> @kpajdzik If I export these classes into the public interface of the library, would that entail a minor version bump? I assume this library follows SemVer 2.0.

Correct!

daviwil:
OK, after looking over the code in `ms-rest-js` and `ms-rest-nodeauth` and thinking about it for a while, here’s why I think both of these classes should live in `ms-rest-nodeauth` inside of the `identity` folder:

1.	The goal of the new "Azure Identity" effort is to produce an authentication library which can pull credentials from the deployment environment so that a developer does not need to write their own credential lookup logic.
2.	In the JavaScript space, the only place where this type of environment credential probing logic makes sense (for now) is in server-side Node.js code deployed to Azure.
3.	Since the actual implementation(s) of `TokenCredential` which handle this credential probing logic will be implemented in `ms-rest-nodeauth`, the two classes in this PR would need to be exported from `ms-rest-js` so that they could be used in this module.  `BearerTokenAuthenticationPolicy` directly depends on `TokenCredential` so they will need to live in the same library, therefore I think their home should be `ms-rest-nodeauth` for the aforementioned reasons.
4.	This new credential probing logic can be an internal implementation detail in `ms-rest-nodeauth` for now, so there’s not yet a need to export these classes in the module’s public interface, hence `TokenCredential` shouldn’t cause a conceptual conflict to users of `ms-rest-nodeauth`.  The only difference to an outside consumer could be a new authentication method that can be used to invoke this credential probing logic but it shouldn’t need to reference these types directly (please correct me if I’m wrong!).
5.	Once the official “Azure Identity” TS module is produced, we can move these classes there and expose them from its public interface.  `ms-rest-*` modules could then take a dependency on that new module to maintain the previous credential probing behavior.

How does that sound to everyone?

daviwil:
Hey folks, after discussing this a bit more with @AlexGhiondea and @bterlson, I've decided to close this PR and add these classes to a different module.  I'll link the new PR once it's up.

PR #62, opened 2019-05-23
get subscriptions if the tokenAudience is for Resource Manager

### LINKED ISSUE #60: Subscriptions not returned calling loginWithServicePrincipalSecretWithAuthResponse when tokenAudience is 'RoleBased'
**Package Version**: 2.0.0
- [ ] **nodejs**
   - **nodejs version**: 8.10.0
   - **os name/version**: Fedora 29

**Describe the bug**
To be honest, I'm not sure if this is a bug or intended change in behavior between the older `ms-rest-azure` and this newer package. As part of my application logic, I perform some checks around the subscriptions associated with the principal that authenticated. When using the `ms-rest-azure`, authentication and getting the list of associated subscriptions could be accomplished by using the `loginWithServicePrincipalSecretWithAuthResponse` function. 

Since the `ms-rest-azure` package is on a path to deprecation, I upgraded my application to use this library instead which has the same function. However, in any authentication response the list of subscriptions returned is always empty. I checked the source code and there appears to be a difference in logic on whether the `subscriptions` property should be populated:

ms-rest-azure: https://github.com/Azure/azure-sdk-for-node/blob/8691046c2ca738552d815f6967f5f54b04840276/runtime/ms-rest-azure/lib/login.js#L440
ms-rest-nodeauth: https://github.com/Azure/ms-rest-nodeauth/blob/master/lib/login.ts#L165

**To Reproduce**
I have a gist with a short script to reproduce this behavior: https://gist.github.com/dwalleck/4d049d3515e82bb4b687fe95a1da194d

1. Call the `loginWithServicePrincipalSecretWithAuthResponse` function of `@azure/ms-rest-nodeauth` with service principal credentials.
2. The `subscriptions` property is an empty list

**Expected behavior**
The `subscriptions` property should be propagated with any subscriptions associated with the user.

**Additional context**

The output I'm currently getting from my test script shows the difference in behavior:

```
msRestAzure subscriptions: [
    {
        "id": "XXXXX",
        "state": "Enabled",
        "authorizationSource": "RoleBased",
        "tenantId": "XXXXX",
        "user": {
            "name": "XXXXX",
            "type": "servicePrincipal"
        },
        "environmentName": "Azure",
        "name": "XXXXX"
    }
]
msRestNodeAuth subscriptions: []
```

amarzavery:
fixes #60 .


lib/login.ts
```diff
@@ -17,6 +17,23 @@ import { MSIVmTokenCredentials, MSIVmOptions } from "./credentials/msiVmTokenCre
 import { MSIAppServiceTokenCredentials, MSIAppServiceOptions } from "./credentials/msiAppServiceTokenCredentials";
 import { MSITokenResponse } from "./credentials/msiTokenCredentials";
 
+/**
+ * @constant {Array<string>} managementPlaneTokenAudiences - Urls for management plane token
+ * audience across different azure environments.
+ */
+const managementPlaneTokenAudiences = [
+  "https://management.core.windows.net/",
```
kpajdzik: Shouldn't that be in `@azure/ms-rest-env`?

lib/login.ts
```diff
@@ -162,10 +179,8 @@ export async function withUsernamePasswordWithAuthResponse(username: string, pas
     await creds.getToken();
     // The token cache gets propulated for all the tenants as a part of building the tenantList.
     tenantList = await buildTenantList(creds);
-    // We only need to get the subscriptionList if the tokenAudience is for a management client.
-    if (options.tokenAudience && options.tokenAudience === options.environment.activeDirectoryResourceId) {
-      subscriptionList = await getSubscriptionsFromTenants(creds, tenantList);
-    }
+    subscriptionList = await _getSubscriptions(creds, tenantList, options.tokenAudience);
```
kpajdzik: @daschult mentioned we don't really use `_` convention. Can we be consistent with other projects?

kpajdzik:
I didn't see it was already merged! :D

PR #64, opened 2019-06-12
Added a new DomainCredentials class for make it easier to supply EventGrid domain credentials

kalyanaj:
Added a new DomainCredentials class for make it easier to supply EventGrid domain credentials. Event Grid domain is a recently introduced concept/resource and applications can publish events to an EventGrid domain for which an api key needs to be provided, and this newly introduced DomainCredentials class makes it easy to do so.


amarzavery:
Please update the version in package.json and update the Changelog.md with the changes for the new version.

kalyanaj:
Thanks @amarzavery for the review!  I have incorporated the changes you suggested, can you please take a look when you get a chance?

amarzavery:
Closing this. The credentials object should only be in ms-rest-js package. we shall be removing the other EG credentials from here. By putting it in ms-rest-js they can be used in node.js and browser.

PR #65, opened 2019-06-13
Enforce azure cli to output json

wilsonge:
If you've run `az configure` to get non-json default output then you don't get JSON. As a result the `JSON.parse` immediately fails due to trying to parse the non-json format. This enforces retrieving in JSON format so that we can always parse it whatever the users default value


amarzavery:
Thank you for contributing the project. The PR has been merged and the new version of the package will be published to npm in ~10 mins.

PR #66, opened 2019-07-19
Make token credentials compatible with @azure/core-auth's TokenCredential

daviwil:
This change makes `ms-rest-nodeauth` token credential implementations compatible with `@azure/core-auth`'s `TokenCredential`, enabling them to be used with Track 2 SDK libraries.  This is part of the resolution for https://github.com/Azure/azure-sdk-for-js/issues/4108.

I've verified these changes with `DeviceTokenCredentials`, `AzureCliCredentials`, and `ApplicationTokenCredentials` using a version of `@azure/arm-resources` that does not accept `ServiceClientCredentials` in its `ServiceClient` constructor.  I also removed the use of `signingPolicy` in `@azure/core-auth` in my local build just to make sure these credentials are being used with `BearerTokenAuthenticationPolicy`.

One implication of this change is that the fix we made to `isTokenCredential` in https://github.com/Azure/azure-sdk-for-js/pull/4335 now has to be rolled back because all of these credentials now expose both `getToken` and `signRequest` due to implementing both `ServiceClientCredentials` and `TokenCredential`.  I'll make a PR to `@azure/core-http` if we decide we like this compatibility approach.

**Questions for Reviewers**

- I've made this a major version bump, but do we think the API changes would actually be considered breaking changes since they are only additive?
- What do you think about the name `prepareToken`?  Doesn't seem explicit enough to me so I'm open to ideas for a better name.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,9 @@
 # Changelog
+
+## 3.0.0 - 2019/07/23
+
+- Token credential types are now compatible with the `TokenCredential` interface from [`@azure/core-auth`](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-auth), enabling them to be used with newer SDK libraries in `azure-sdk-for-js`.
+
 ## 2.0.3 - 2019/07/23
 - Updated min version of dependency `@azure/ms-rest-js` to `^2.0.3`.
 - Updated min version of dependenct `@azure/ms-rest-azure-env` to `^2.0.0`.
@@ -13,7 +18,7 @@
 
 ## 2.0.0 - 2019/05/20
 - Added support for client_id, object_id and ms_res_id query parameters for VmMSI. Fixes [#58](https://github.com/Azure/ms-rest-nodeauth/issues/58).
-- **Breaking change:** 
+- **Breaking change:**
   - Added support to get token for a different resource like Azure Keyvault, Azure Batch, Azure Graph apart from the default Azure Resource Manager resource via `AzureCliCredentials`.
   - `AzureCliCredentials.create()` now takes an optional parameter where the user can specify the subscriptionId and the resource for which the token is required.
   - `AzureCliCredentials.getDefaultSubscription()` has been changed to `AzureCliCredentials.getSubscription(subscriptionIdOrName?: string)`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 3.0.0 - 2019/07/23
+## 3.0.0 - 2019/07/26
 
 - Token credential types are now compatible with the `TokenCredential` interface from [`@azure/core-auth`](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-auth), enabling them to be used with newer SDK libraries in `azure-sdk-for-js`.
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,13 @@
 # Changelog
 
-## 3.0.0 - 2019/07/26
+## 3.0.0 - 2019/08/02
 
 - Token credential types are now compatible with the `TokenCredential` interface from [`@azure/core-auth`](https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/core/core-auth), enabling them to be used with newer SDK libraries in `azure-sdk-for-js`.
+- **Breaking change:**
+    - Updated min version of dependency `@azure/ms-rest-js` to `^2.0.4`.
+
+## 2.0.4 - 2019/08/02
+- Rolling back the min version of dependency `@azure/ms-rest-js` from `^2.0.3` to `^1.8.13` thereby fixing [#69](https://github.com/Azure/ms-rest-nodeauth/issues/69).
 
 ## 2.0.3 - 2019/07/23
 - Updated min version of dependency `@azure/ms-rest-js` to `^2.0.3`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,10 @@
 # Changelog
+## 2.0.3 - 2019/07/23
+- Updated min version of dependency `@azure/ms-rest-js` to `^2.0.3`.
+- Updated min version of dependenct `@azure/ms-rest-azure-env` to `^2.0.0`.
+- Improved documentation of `MSIOptions.resource`
+- Improved samples in README.md
+
 ## 2.0.2 - 2019/06/13
  - Ensure we always get JSON responses back from Azure CLI.
 
--- a/lib/credentials/applicationTokenCertificateCredentials.ts
+++ b/lib/credentials/applicationTokenCertificateCredentials.ts
@@ -3,6 +3,8 @@
 
 import { readFileSync } from "fs";
 import { createHash } from "crypto";
+import { prepareToken } from "./coreAuthHelpers";
+import { AccessToken, GetTokenOptions } from "@azure/core-auth";
 import { ApplicationTokenCredentialsBase } from "./applicationTokenCredentialsBase";
 import { Environment } from "@azure/ms-rest-azure-env";
 import { AuthConstants, TokenAudience } from "../util/authConstants";
@@ -52,10 +54,12 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
    * Tries to get the token from cache initially. If that is unsuccessfull then it tries to get the token from ADAL.
    * @returns {Promise<TokenResponse>} A promise that resolves to TokenResponse and rejects with an Error.
    */
-  public async getToken(): Promise<TokenResponse> {
+  public async getToken(): Promise<TokenResponse>;
+  public async getToken(scopes: string | string[], options?: GetTokenOptions): Promise<AccessToken>;
+  public async getToken(scopes?: string | string[]): Promise<TokenResponse | AccessToken> {
     try {
       const tokenResponse = await this.getTokenFromCache();
-      return tokenResponse;
+      return prepareToken(tokenResponse, scopes);
     } catch (error) {
       if (error.message.startsWith(AuthConstants.SDK_INTERNAL_ERROR)) {
         return Promise.reject(error);
@@ -74,7 +78,7 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
             if (tokenResponse.error || tokenResponse.errorDescription) {
               return reject(tokenResponse);
             }
-            return resolve(tokenResponse as TokenResponse);
+            return resolve(prepareToken(tokenResponse as TokenResponse, scopes));
           }
         );
       });
--- a/lib/credentials/applicationTokenCertificateCredentials.ts
+++ b/lib/credentials/applicationTokenCertificateCredentials.ts
@@ -59,7 +59,8 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
   public async getToken(scopes?: string | string[]): Promise<TokenResponse | AccessToken> {
     try {
       const tokenResponse = await this.getTokenFromCache();
-      return prepareToken(tokenResponse, scopes);
+      const token = prepareToken(tokenResponse, scopes);
+      return token;
     } catch (error) {
       if (error.message.startsWith(AuthConstants.SDK_INTERNAL_ERROR)) {
         return Promise.reject(error);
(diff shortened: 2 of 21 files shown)


daviwil:
Worth noting that the CI build will fail because we haven't shipped `@azure/core-auth` yet.  Once we decide whether the fix in https://github.com/Azure/azure-sdk-for-js/pull/4335 should be rolled back, I'll get that change in and then have the engineering systems team ship that library for us.

lib/credentials/coreAuthHelpers.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Prepares a TokenResponse to be returned as a TokenResponse or
+ * a @azure/core-auth AccessToken depending on whether the 'scopes'
+ * parameter is null or not (the key to determining which getToken
+ * method has been called).
+ */
+export function prepareToken<T extends TokenResponseLike>(
+  token: T,
+  scopes: string | string[] | undefined): T | AccessToken {
+  if (scopes) {
```
amarzavery: What if `scopes` is an empty string ? This check will fail for that. If the intention is to check specifically for `null` or `undefined` then using the "double equal to" or "not with a single equal to" will be a good check as that will take care of `null` and `undefined`. 
` if (scopes != undefined)`

daviwil:
PR Azure/azure-sdk-for-js#4367 restricts the `ServiceClient` interface to only accept `TokenCredential` implementations.

package.json
```diff
@@ -28,6 +28,7 @@
     "tsconfig.json"
   ],
   "dependencies": {
+    "@azure/core-auth": "1.0.0-preview.1",
     "@azure/ms-rest-azure-env": "^1.1.2",
     "@azure/ms-rest-js": "^1.8.7",
```
amarzavery: Should this be replaced by `@azure/core-http`?

lib/credentials/coreAuthHelpers.ts
daviwil: Thanks!  It shouldn't be an empty string, but you never know what a user might do.  I'll make the check stricter.

package.json
daviwil: Nah, we pulled the `TokenCredential` interface and related types into `@azure/core-auth` so that they can be reused in other places.  `@azure/core-http` depends on `@azure/core-auth` in the unshipped code in `master`.

amarzavery:
For safety, I would let this be a major version bump (3.0.0). It keeps a clear boundary between new stuff and the old stuff. Makes things easier for us to maintain.

lib/credentials/coreAuthHelpers.ts
```diff
@@ -0,0 +1,28 @@
...
+    return {
+      token: token.accessToken,
+      expiresOnTimestamp:
+        typeof token.expiresOn === "string"
```
ramya-rao-a: This assumes that `expiresOn` exists on all the credential types in ms-rest-nodeauth.
Are we sure about this assumption?

daviwil: Good point, I can't find `expiresOn` being assigned anywhere in a few of the credential implementations.  @amarzavery, do you remember how users are supposed to deal with token expiration in this library?  I only see `AzureCliCredentials` implementing any kind of token refresh logic.

lib/credentials/coreAuthHelpers.ts
```diff
@@ -0,0 +1,28 @@
+import { AccessToken } from "@azure/core-auth";
+
+export interface TokenResponseLike {
```
ramya-rao-a: Does this have to be exported? I don't see it being referenced anywhere other than this file

lib/credentials/coreAuthHelpers.ts
amarzavery: adal-node has an in memory token cache. It gets a new token if it is about to get expired or gives the token from the token cache.

For AzureCliCredentials we are calling the cli to authenticate and give us the new token that will be valid for the next one hour. Cli knows how to refresh the token if it is using it internally. We can't be calling the Cli for every request, as that will terribly slow down the sdk (spawning a new process and asking the Cli for creds). 

Thus we try to do the same thing that adal-python would do anyways (if we are within the token refresh window (last 5 mins before the token expiry) then ask the Cli for token, it will ask adal-python which will then get the new  token that will be valid for the next one hour).

Thus we need to check the expiresOn property.

lib/credentials/applicationTokenCertificateCredentials.ts
```diff
@@ -74,7 +78,7 @@ export class ApplicationTokenCertificateCredentials extends ApplicationTokenCred
             if (tokenResponse.error || tokenResponse.errorDescription) {
               return reject(tokenResponse);
             }
-            return resolve(tokenResponse as TokenResponse);
+            return resolve(prepareToken(tokenResponse as TokenResponse, scopes));
```
ramya-rao-a: Nit: Where-ever applicable, please consider having a const for the prepared token instead of inlining the call to `prepareToken` in this manner. Having the call in a separate line would make it easier to add breakpoints when debugging

test/credentials/coreAuthHelpersTests.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+import { assert } from "chai";
+import { prepareToken } from "../../lib/credentials/coreAuthHelpers";
+
+describe("prepareToken", function() {
```
ramya-rao-a: There are 2 test files for MSI creds at the moment.
Can we update the test cases in these files to include a call to getToken(scopes) to check that the returned value is `AccessToken`?

ramya-rao-a:
>Once we decide whether the fix in Azure/azure-sdk-for-js#4335 should be rolled back ...

Having creds from ms-rest-nodeauth be compatible with core-auth doesnt solve the problem that was being fixed in https://github.com/Azure/azure-sdk-for-js/pull/4335 does it?

Or are we expecting that a combination of the above and https://github.com/Azure/azure-sdk-for-js/pull/4367, solves the issue in  https://github.com/Azure/azure-sdk-for-js/pull/4335

lib/credentials/coreAuthHelpers.ts
daviwil: Possibly not, I'll give it a shot.

lib/credentials/applicationTokenCertificateCredentials.ts
daviwil: Sounds good, I'll make that change.

test/credentials/coreAuthHelpersTests.ts
daviwil: I'll give it a shot

daviwil:
> Or are we expecting that a combination of the above and https://github.com/Azure/azure-sdk-for-js/pull/4367, solves the issue in  https://github.com/Azure/azure-sdk-for-js/pull/4335 

Yep, it's the combination of fixes that mostly solves the problem.  Previously the problem was that `ms-rest-nodeauth` credentials implemented `TokenClientCredentials` which has its own `getToken` method that takes no parameters.  `isTokenCredential` would incorrectly identify such a credential as `TokenCredential` and then use `BearerTokenAuthenticationPolicy` for it instead of `SigningPolicy`.  This would mean that the credential from `ms-rest-nodeauth` would get used incorrectly and probably throw an error (or at the very least authentication would fail).

After these new two PRs, we only ever use `BearerTokenAuthenticationPolicy` inside of `ServiceClient` so the `isTokenCredential` check doesn't need to be as strict.  In fact, the check for `signRequest` now becomes a problem for `ms-rest-nodeauth` credentials because they would not be evaluated as `TokenCredential` implementations anymore due to having a `signRequest` method (even though they also implement `TokenCredential`).  Thus the check had to be removed.

This does leave us with one problem, though: we don't have a reliable way to distinguish between a `TokenClientCredentials` and a `TokenCredential` because they both have a `getToken` method, just with different parameters and return type.  As @bterlson mentioned in Azure/azure-sdk-for-js#4335, using `getToken.length` to get the parameter count is unreliable, so at the moment there isn't a better way to distinguish between the two interfaces.

I'd be happy to hear if someone has another idea though!

daviwil:
Had a better idea on how to make `isTokenCredential` a bit more reliable, left it as a comment over on the other PR: https://github.com/Azure/azure-sdk-for-js/pull/4367#issuecomment-513902599

lib/credentials/coreAuthHelpers.ts
daviwil: On second thought, the idea is check for whether a parameter was passed to `scopes`, not to scrutinize whether the value of `scopes` is valid.  I think in this case I'm only concerned about whether it's explicitly `undefined` since that's what the parameter would contain if nothing was passed.

`BearerTokenAuthenticationPolicy` in `core-http` will always pass a string or an array to the `scopes` parameter so I think it's pretty safe to only check for `undefined` since we don't actually use the real value of `scopes` in these forward-compatible implementations anyway.

test/credentials/coreAuthHelpersTests.ts
daviwil: Added a test in each MSI spec file to verify that an `AccessToken` is returned when a scope is passed to `getToken`.  Is that sufficient?

lib/credentials/applicationTokenCertificateCredentials.ts
daviwil: Done!

lib/credentials/coreAuthHelpers.ts
daviwil: Looks like removing the export worked fine!

lib/credentials/coreAuthHelpers.ts
daviwil: I've loosened the expectation of an `expiresOn` property on the token and put a default expiration timestamp of an hour if one is not provided.  What do you think?

lib/credentials/azureCliCredentials.ts
```diff
@@ -120,7 +122,9 @@ export class AzureCliCredentials implements TokenClientCredentials {
    * changed else uses the cached accessToken.
    * @return The tokenResponse (tokenType and accessToken are the two important properties).
    */
-  public async getToken(): Promise<TokenResponse> {
+  public getToken(): Promise<TokenResponse>;
+  public getToken(scopes: string | string[], options?: GetTokenOptions): Promise<AccessToken>;
```
ramya-rao-a: Why no `async` keywords for the overloads?

lib/credentials/coreAuthHelpers.ts
ramya-rao-a: Did you come across any of the credentials using the `expiresIn` property?

lib/credentials/deviceTokenCredentials.ts
```diff
@@ -53,8 +55,11 @@ export class DeviceTokenCredentials extends TokenCredentialsBase {
     this.username = username;
   }
 
-  public getToken(): Promise<TokenResponse> {
+  public getToken(): Promise<TokenResponse>;
+  public getToken(scopes: string | string[], options?: GetTokenOptions): Promise<AccessToken>;
+  public async getToken(scopes?: string | string[]): Promise<TokenResponse | AccessToken> {
```
ramya-rao-a: Missing async keywords on the overloads

Changelog.md
```diff
@@ -1,4 +1,9 @@
 # Changelog
+
+## 3.0.0 - 2019/07/19
```
ramya-rao-a: Date would need an update

lib/credentials/azureCliCredentials.ts
daviwil: Because the overload definitions are just typings and `async` is only really needed with a function body that uses `await`.  That said, I'll add the keywords to minimize confusion.

Changelog.md
daviwil: I'll update it before merging

lib/credentials/azureCliCredentials.ts
ramya-rao-a: My confusion was because in a couple of places we did add the `async` keyword for the overloads, and in some places we didnt

daviwil:
This PR is no longer necessary as we've made `@azure/core-http` accept both `ServiceClientCredentials` and `TokenCredential`

PR #70, opened 2019-08-02
fix #69 by rolling back @azure/ms-rest-js to ^1.8.13

### LINKED ISSUE #69: Breaking change in a patch version
**Package Version**: 2.0.3
- [x] **nodejs**
   - **nodejs version**: 10.15.3
   - **os name/version**: Windows 10

**Describe the bug**
I'm getting #67 but the other way around. My app depends on 

```
"@azure/ms-rest-azure-js":       "~1.3.8",
"@azure/ms-rest-nodeauth":       "~2.0.2",
```

Both used to be the latest versions of corresponding libraries not so long ago. The latest patch version `2.0.3` of `ms-rest-nodeauth` updated to `ms-rest-js` version `2.0.3` which is not compatible to the old `ms-rest-js` referenced by the old `ms-rest-azure-js`. Thus, TypeScript throws `...is not assignable to type...` errors.

While version reconciliation is fine by itself, technically it's a breaking change, so I wouldn't expect it to happen in a patch version.

**Expected behavior**

Don't make breaking changes in patch versions.

amarzavery:
- fixes #69

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,7 @@
 # Changelog
+## 2.0.3 - 2019/08/02
+- Rolling back the min version of dependency `@azure/ms-rest-js` from `^2.0.3` to `^1.8.13` thereby fixing [#69](https://github.com/Azure/ms-rest-nodeauth/issues/69).
+
 ## 2.0.3 - 2019/07/23
 - Updated min version of dependency `@azure/ms-rest-js` to `^2.0.3`.
 - Updated min version of dependenct `@azure/ms-rest-azure-env` to `^2.0.0`.
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,5 @@
 # Changelog
-## 2.0.3 - 2019/08/02
+## 2.0.4 - 2019/08/02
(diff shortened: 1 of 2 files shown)


Changelog.md
```diff
@@ -1,4 +1,7 @@
 # Changelog
+## 2.0.4 - 2019/08/02
+- Rolling back the min version of dependency `@azure/ms-rest-js` from `^2.0.3` to `^1.8.13` thereby fixing [#69](https://github.com/Azure/ms-rest-nodeauth/issues/69).
```
ramya-rao-a: What about azure/ms-rest-azure-env? The breaking changes there from major version 1 to 2 doesn't affect anybody?

amarzavery: That shouldn't affect anyone. It is a good breaking change. We removed  a dot from the storage suffix which shouldn't have been there in the first place. This change was made after a customer filed an issue. Since it was a breaking change we updated the major version of the package. However, it is a good change and we don't expect it to cause issues to our customers since the previous code was not usable.

PR #74, opened 2019-08-06
Cleanup `async function`s and `Promise` based code

arthurschreiber:
This cleans up a bunch of misunderstandings between `async` functions and `Promise`s.

* There is no need to catch exceptions inside async functions and return them wrapped via `Promise.reject`. This will happen automatically if an exception is thrown inside an async function.

* There is no need to wrap return values of async functions with `Promise.resolve`. This will happen automatically.

These changes make the code easier to read and understand, and remove unneccessary overhead.

There's also a few other commits in here that clean up some of the code.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,8 @@
 # Changelog
+
+## 3.0.1 - 2019/08/06
+- Reduce number of `Promise` object allocations inside `async` functions.
+
 ## 3.0.0 - 2019/08/02
 - **Breaking change:**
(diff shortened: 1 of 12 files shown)


amarzavery:
@arthurschreiber - Thank you for your contribution. The changes LGTM.

- Please update the patch version in package.json by `1` -> `3.0.1`.
- Please update the changelog.md with a brief summary of the changes made in this PR.

arthurschreiber:
@amarzavery Done! 👍

PR #75, opened 2019-08-07
Fix a bunch of type related issues

arthurschreiber:
This fixes a whole bunch of type related issues in this library.

I tried to make the commits self-contained.

I'll add some comments explaining some of the changes that I made. 🙇

### DIFF
--- a/lib/login.ts
+++ b/lib/login.ts
@@ -156,9 +156,9 @@ export type Callback<TResult> = (error?: Error, result?: TResult) => void;
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse<UserTokenCredentials>>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
  */
-export async function withUsernamePasswordWithAuthResponse(username: string, password: string, options?: LoginWithUsernamePasswordOptions): Promise<AuthResponse> {
+export async function withUsernamePasswordWithAuthResponse(username: string, password: string, options?: LoginWithUsernamePasswordOptions): Promise<AuthResponse<UserTokenCredentials>> {
   if (!options) {
     options = {};
   }
@@ -196,9 +196,9 @@ export async function withUsernamePasswordWithAuthResponse(username: string, pas
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse<ApplicationTokenCredentials>>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
  */
-export async function withServicePrincipalSecretWithAuthResponse(clientId: string, secret: string, domain: string, options?: AzureTokenCredentialsOptions): Promise<AuthResponse> {
+export async function withServicePrincipalSecretWithAuthResponse(clientId: string, secret: string, domain: string, options?: AzureTokenCredentialsOptions): Promise<AuthResponse<ApplicationTokenCredentials>> {
   if (!options) {
     options = {};
   }
@@ -230,9 +230,9 @@ export async function withServicePrincipalSecretWithAuthResponse(clientId: strin
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse<ApplicationTokenCertificateCredentials>>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
  */
-export async function withServicePrincipalCertificateWithAuthResponse(clientId: string, certificateStringOrFilePath: string, domain: string, options?: AzureTokenCredentialsOptions): Promise<AuthResponse> {
+export async function withServicePrincipalCertificateWithAuthResponse(clientId: string, certificateStringOrFilePath: string, domain: string, options?: AzureTokenCredentialsOptions): Promise<AuthResponse<ApplicationTokenCertificateCredentials>> {
   if (!options) {
     options = {};
   }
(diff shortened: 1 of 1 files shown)


lib/login.ts
```diff
@@ -106,11 +105,11 @@ export interface InteractiveLoginOptions extends LoginWithUsernamePasswordOption
 /**
  * @interface AuthResponse - Describes the authentication response.
  */
-export interface AuthResponse {
+export interface AuthResponse<T extends TokenCredentialsBase = TokenCredentialsBase> {
```
arthurschreiber: Making `AuthResponse` a generic type is backwards compatible as it comes with a default type for the generic parameter, which matches the old type. This should only require a bump of the minor version number. 👍 

Making this type generic allows us to keep the type information for the actual credential type, instead of just the abstract `TokenCredentialsBase`.

The default type probably should be removed in a later, major release.

lib/login.ts
```diff
@@ -601,20 +600,20 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
  */
 export function interactive(): Promise<DeviceTokenCredentials>;
 export function interactive(options: InteractiveLoginOptions): Promise<DeviceTokenCredentials>;
-export function interactive(options: InteractiveLoginOptions, callback: { (err: Error, credentials: DeviceTokenCredentials, subscriptions: Array<LinkedSubscription>): void }): void;
+export function interactive(options: InteractiveLoginOptions, callback: { (err: Error | undefined, credentials?: DeviceTokenCredentials, subscriptions?: Array<LinkedSubscription>): void }): void;
```
arthurschreiber: Here (and in other places), the type information for the callback was not actually mirroring how the callback was called. This was due to the `callback as Function` casting, which is not type safe and did hide this issue.

I'm not sure if this can be seen as a fix, or warrants a major release. 🤷‍♂

lib/login.ts
```diff
@@ -601,20 +600,20 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
...
 export function interactive(callback: any): void;
-export function interactive(options?: InteractiveLoginOptions, callback?: { (err: Error, credentials: DeviceTokenCredentials, subscriptions: Array<LinkedSubscription>): void }): any {
+export function interactive(options?: InteractiveLoginOptions, callback?: { (err: Error | undefined, credentials?: DeviceTokenCredentials, subscriptions?: Array<LinkedSubscription>): void }): void | Promise<DeviceTokenCredentials> {
```
arthurschreiber: Here (and in other places), I made the return type definition for the implementation function stricter. This helps catching type specific bugs earlier. As the implementation function's type is never exposed to users, this change is backwards compatible. 👍

lib/login.ts
```diff
@@ -601,20 +600,20 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
...
     options = undefined;
   }
-  const cb = callback as Function;
   if (!callback) {
     return withInteractiveWithAuthResponse(options).then((authRes) => {
       return authRes.credentials;
     });
   } else {
-    msRest.promiseToCallback(withInteractiveWithAuthResponse(options))((err: Error, authRes: AuthResponse) => {
+    const cb = callback;
```
arthurschreiber: We don't need any casting of `callback` here, as typescript can figures out itself that `callback` is not undefined here. We still need to assign to a different variable as typescript otherwise complains that `callback` could be undefined in the callback function passed to `msRest.promiseToCallback`.

lib/login.ts
```diff
@@ -700,21 +699,21 @@ export function withServicePrincipalSecret(clientId: string, secret: string, dom
  *             @reject {Error} - The error object.
  */
 export function withServicePrincipalCertificate(clientId: string, certificateStringOrFilePath: string, domain: string): Promise<ApplicationTokenCertificateCredentials>;
-export function withServicePrincipalCertificate(clientId: string, certificateStringOrFilePath: string, domain: string, options: AzureTokenCredentialsOptions): Promise<ApplicationTokenCredentials>;
-export function withServicePrincipalCertificate(clientId: string, certificateStringOrFilePath: string, domain: string, options: AzureTokenCredentialsOptions, callback: { (err: Error, credentials: ApplicationTokenCertificateCredentials, subscriptions: Array<LinkedSubscription>): void }): void;
+export function withServicePrincipalCertificate(clientId: string, certificateStringOrFilePath: string, domain: string, options: AzureTokenCredentialsOptions): Promise<ApplicationTokenCertificateCredentials>;
```
arthurschreiber: Here, the type was just wrong. `ApplicationTokenCredentials` and `ApplicationTokenCertificateCredentials` are not interchangeable. 🤷‍♂

Again, not sure if that's considered just a fix or requires a major version bump.

lib/login.ts
```diff
@@ -853,15 +852,15 @@ export function loginWithVmMSI(options?: MSIVmOptions | Callback<MSIVmTokenCrede
     callback = options;
     options = {};
   }
-  const cb = callback as Function;
   if (!callback) {
     return _withMSI(options as MSIVmOptions);
   } else {
-    msRest.promiseToCallback(_withMSI(options as MSIVmOptions))((err: Error, tokenRes: MSITokenResponse) => {
+    const cb = callback;
+    msRest.promiseToCallback(_withMSI(options as MSIVmOptions))((err: Error, creds: MSIVmTokenCredentials) => {
```
arthurschreiber: `msRest.promiseToCallback` is problematic as it erases types (everything gets casted to `any` and then needs to be casted to a specific type manually). This hides typing issues, as can be seen here, were a completely incorrect type was used. 😞

This could be fixed by either fixing the type declaration for `msRest.promiseToCallback` (but that's located in a different project), or by moving away from that helper. In the meantime, this change fixes the issue.

daviwil:
Hi @arthurschreiber, thanks for sending this great PR!

We're currently in the process of winding down development on this library in favor of the new `@azure/identity` library that's being developed for our [next wave of TypeScript SDKs](https://azure.microsoft.com/en-us/blog/previewing-azure-sdks-following-new-azure-sdk-api-standards/) so we are less likely to accept PRs for `ms-rest-nodeauth` at this point.  Were any of these changes fixes for actual bugs that blocked your ability to use the library?

Out of curiosity, where have you been using `ms-rest-nodeauth` so far?

arthurschreiber:
@daviwil Oh wow, I did not know this library is getting deprecated. 🙇 

I'm one of the maintainers of [`tedious`](https://github.com/tediousjs/tedious), the Microsoft recommended Node.js package for connecting to SQL Server and Azure SQL databases. Azure AD and MSI authentication support was implemented via `@azure/ms-rest-nodeauth`. And the fixes I've provided so far were all things I've found that didn't seem right.

I guess migrating to `@azure/identity` would probably be a better use of my time? 😅

daviwil:
It's slowly being deprecated, but yeah, it'd be better to focus on `@azure/identity`.  How are you currently using these credentials in `tedious`?  I could give you some pointers on how you might be able to switch over to `@azure/identity`, I don't think it'd be a significant amount of work depending on what you're doing currently.

arthurschreiber:
@daviwil https://github.com/tediousjs/tedious/blob/7721dac034ded633e9fa8d85155a311243fbdd45/src/connection.js#L2076-L2117

Yeah, I took a quick glance over `@azure/identity`, and it should be pretty straightforward to switch. 👍

daviwil:
Please let me know if you have any questions while trying it out!  That library is still in preview so your feedback will be very valuable.

funkydev:
@daviwil I have a question about `InteractiveBrowserCredential` in node.js app.

This authorization method is supported by the current library but by `@azure/identity` is not. Is there any info about implementation of the interactive authentication?

daviwil:
@funkydev as of `@azure/identity` `1.0.0-preview.2`, we now provide an [`InteractiveBrowserCredential`](https://azure.github.io/azure-sdk-for-js/identity/classes/interactivebrowsercredential.html) for the in-browser interactive auth flow, but we don't have such an implementation for Node.js yet.  We've got it on our internal roadmap for a future preview but we don't know when it will arrive.

funkydev:
@daviwil So I have opened the issue [#4774](https://github.com/Azure/azure-sdk-for-js/issues/4774) which requests that. 

People can't move to `@azure/identity` library, when it doesn't support the same functionality. 
`tedious` library needs support for Interactive login.

daviwil:
Thank you for filing that! We will let you know when that gets added.

arthurschreiber:
@ramya-rao-a Based on https://github.com/tediousjs/tedious/issues/1146#issuecomment-696241577, would it make sense to get these changes merged and the type related issues resolved? 🙇‍♂️

ramya-rao-a:
Hey @arthurschreiber 

Thanks for your patience here

Would you consider splitting this into two separate PRs?
- One can be for the AuthResponse<T> update and all its resulting changes. This would be pretty straight forward for me to review, approve and merge right away
- The other for all the changes around the callback. We need to dig a little deeper here

ramya-rao-a:
@arthurschreiber Created #121 for the AuthResponse type changes

ramya-rao-a:
Thanks for the work here @arthurschreiber 

I have pulled out the changes for strengthening the types around AuthResponse in #121 which is now merged and an update will be released soon

Closing this PR as we havent seen much activity here. If there are any more improvements that can be made in a non breaking way, please send a new PR

PR #76, opened 2019-08-16
Fix bug preventing any tenants from being discovered

weswigham:
Which in turns prevents any subscriptions in turn from being discovered which in turn makes auth very not useful.


amarzavery:
update the version in package.json and the changelog.

PR #77, opened 2019-08-22
Fix callback not calling issue in applicationTokenCredentialsBase.ts

XinxingLiu:
When calling "loginWithServicePrincipalSecretWithAuthResponse", the callback function is sometimes not called because a nested promise with parameter "resolve" hides the same parameter name in outer promise in applicationTokenCredentialsBase.ts. 

This bug is blocking our autorest generated SDK (typescript code) to run. Since the generated code depends on ms-rest-js@1.8.1, we cannot consume ms-rest-nodeauth@3.0.0 or higher. The fix should be based on ms-rest-nodeauth@2.0.4 and probably we need a new package with version 2.0.5 published in npm public server, to get our code working.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.2 - 2019/08/22
+
+- Fixed a bug where the callback to `loginWithServicePrincipalSecretWithAuthResponse` is sometimes not called.
+
 ## 3.0.2 - 2019/08/16
 - Fix bug prevent tenant IDs from being discovered on auth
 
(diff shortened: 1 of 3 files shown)


amarzavery:
The changes LGTM. /cc @daviwil - Please take a look at this. 
@XinxingLiu - Some questions:
 - Which version of `@microsoft.azure/autorest.typescript` generator are you using?
 - For which service are you generating the SDK? Is there a swagger spec for your service in the https://github.com/azure/azure-rest-api-specs repo? If yes, then can you point to the location of the swagger spec?

ramya-rao-a:
@amarzavery This is a fix we will have to take both for v3 and v2 of ms-rest-nodeauth

I have directed @XinxingLiu to use the latest version of the generator so that the latest ms-rest-nodeuath can be used.

amarzavery:
@ramya-rao-a - That sounds good! Please make sure to update the package.json version and the changelog accordingly.

daviwil:
I will help get the version and changelog updated for `master` and then after merging I'll set up the necessary branch and pipeline to ship an update to the 2.x series.

daviwil:
Version 3.0.3 has been published, now working on getting version 2.0.5 published.

daviwil:
Version 2.0.5 is out: https://www.npmjs.com/package/@azure/ms-rest-nodeauth

@XinxingLiu, let me know if the new update fixes the issue for you!

Issue #88, opened 2020-05-01
Samples for getting MSI from App Service should use IDENTITY_ENDPOINT


dpolivy:
This is a doc issue, not a code issue.

The sample showing how to [login using an MSI from App Service](https://github.com/Azure/ms-rest-nodeauth#msi-managed-service-identity-based-login-from-an-appservice-or-azure-function-created-in-azure) contains the following snippet:
````
const options: msRestNodeAuth.MSIAppServiceOptions = {
  msiEndpoint: "http://127.0.0.1:41741/MSI/token/";
}
````

However, per the [App Service documentation](https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=javascript#obtain-tokens-for-azure-resources), one is supposed to use the environment variable `IDENTITY_ENDPOINT` to get the URL for the token service.

It took me a little bit to figure this out. The sample code in the README would be more effective if it used this in the example (and perhaps linked to the App Service page for more details), e.g.:
````
const options: msRestNodeAuth.MSIAppServiceOptions = {
  msiEndpoint: process.env["IDENTITY_ENDPOINT"]
}
````

sadasant:
@dpolivy Thank you, Dan!

We will be improving the documentation based on your feedback. We appreciate your interest in our projects!

dpolivy:
@sadasant Thanks! I just took a look at the linked PR and added one quick comment for you, hope it is useful :) But yes, looks like the changes may address my feedback.

sadasant:
Thank you @dpolivy !

ramya-rao-a:
With the latest version 3.0.5 of the package, `IDENTITY_ENDPOINT` and `IDENTITY_SECRET` environment variables will be respected when logging in using MSI from App Service.

Thanks for reporting @dpolivy

Issue #86, opened 2020-03-22
MSIAppServiceTokenCredentials doesn't support single user-assigned managed identities


wuzhuoqing:
Package Version: _______
nodejs
nodejs version: 3.0.3
Describe the bug
Azure supports two types of managed identities: system-assigned and user-assigned. Currently, MSIAppServiceTokenCredentials doesn't support user-assigned identities.

**To Reproduce**
1. 
Create node.js function app, assign only one user-assigned identity
const msRestNodeAuth = require("@azure/ms-rest-nodeauth");
const creds = await msRestNodeAuth.loginWithAppServiceMSI();

**Expected behavior**
creds return successfully

**Screenshots**
Exception: Error: Invalid token response, did not find tokenType. Response body is: {"statusCode":400,"message":"Unable to load requested managed identity.","correlationId":"d7e4de93-374a-428f-b6c9-21d9c42966f5"}
Stack: Error: Invalid token response, did not find tokenType. Response body is: {"statusCode":400,"message":"Unable to load requested managed identity.","correlationId":"d7e4de93-374a-428f-b6c9-21d9c42966f5"}
    at MSIAppServiceTokenCredentials.<anonymous> (/home/site/wwwroot/node_modules/@azure/ms-rest-nodeauth/dist/lib/credentials/msiAppServiceTokenCredentials.js:70:23)
    at Generator.next (<anonymous>)
    at fulfilled (/home/site/wwwroot/node_modules/@azure/ms-rest-nodeauth/dist/lib/credentials/msiAppServiceTokenCredentials.js:6:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

**Additional context**
This maybe a bug in backend MSI agent where it should return token when only one user-assigned identity is available. However don't know if the backend MSI agent is open source or how to create issue for that.

Note that this is similar to https://github.com/Azure/ms-rest-nodeauth/issues/79, but when there is only one user-assigned identity so the client id should not be needed.

Even though the [doc](https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=javascript) says "If omitted, the system-assigned identity is used." it is not user friendly as the code will need to know what is the only client id and not consistent with vmMSI where it can get token without problem.
creds = await msRestNodeAuth.loginWithVmMSI();

sadasant:
[...]

PR #85, opened 2020-02-29
Add clientid param to MSIAppServiceOptions

emily-curry:
Fixes #79, allows for a clientid param to be provided to `MSIAppServiceOptions` so that code in App Service may generate credentials using a user-assigned managed identity.

Usage of `WebResource.prepare` within `prepareRequestOptions` has been mimicked from [#59](https://github.com/Azure/ms-rest-nodeauth/pull/59/files#diff-ef7041f1621d29386b2a2ed0db058220) for consistency. Structure of headers/params in the request has been based on the [app-service/overview-managed-identity](https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=dotnet#using-the-rest-protocol) page.

PR #91, opened 2020-05-19
[Issue 89] Exposing the buildTenantList method

### LINKED ISSUE #89: No subscriptions found when using personal or trial Azure accounts with interactive login

sadasant:
In this PR, I:

- Made `buildTenantList` public, to allow users to retrieve all of their tenants after authenticating.
- Added a new public method to `TokenCredentialsBase` called `setDomain` that can be used to force Adal to renew the token against a different tenant.
- Added a sample expressing different ways to work around the [issue #89](https://github.com/Azure/ms-rest-nodeauth/issues/89).

---

### Background

Issue [89](https://github.com/Azure/ms-rest-nodeauth/issues/89) shows that `interactiveLoginWithAuthResponse` is not returning any subscription for personal accounts with one or more tenants. This issue is not perceivable until the code explicitly calls to `_getSubscriptions`, which returns an empty array for those cases.

Upon investigation, a possible solution was to call `interactiveLoginWithAuthResponse` again with one of the received tenants from a previous execution of the same function as the `domain` parameter of this new call.

Another alternative seemed to set `organizations` as the domain, which should work in the v2.0 of the OAuth2 authentication workflow, as described in:

> The `{tenant}` value in the path of the request can be used to control who can sign into the application. The allowed values are `common`, `organizations`, `consumers`, and tenant identifiers. For more detail, see protocol basics.
> ([Source](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow))

However, passing in `organizations` as the domain causes the network requests to answer with status code 400. This is likely because `ms-rest-nodeauth` is using v1.0 of the API (`adal-node` specifically uses version 1.0 by default: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/oauth2client.js#L73).

Assuming a change of OAuth API version is out of the table, the repeated authentication approach could be narrowed down to a modification of `withInteractiveWithAuthResponse`, in which after retrieving the tenants, we could re-generate the credentials from `adal-node` to retrieve the subscriptions. This could be done by generating new credentials with each tenant as the domain of the credential. Though this approach works, this is also what our `_getSubscriptions` method is supposed to be doing by altering the credential's domain:

https://github.com/Azure/ms-rest-nodeauth/blob/31285f9e384c73ab5c0179565c328f5534a50679/lib/subscriptionManagement/subscriptionUtils.ts#L112

However, Adal does not use this domain. When we generate a new `TokenCredentialsBase`, we create a new `authorityUrl` based on the domain sent directly to the constructor, from which we generate a new `AuthenticationContext`:

https://github.com/Azure/ms-rest-nodeauth/blob/31285f9e384c73ab5c0179565c328f5534a50679/lib/credentials/tokenCredentialsBase.ts#L34

Adal stores this `authorityUrl` internally, and uses these passed-in values to validate any token in the cache:

- Where Adal uses the authorityUrl: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/authentication-context.js#L92
- Which is stored here: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/authority.js#L41
- What Adal's Authentication Context uses to provide the authority URL to any external entity: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/authentication-context.js#L110
- How Adal uses this local object to validate the cache inside of the Authority methods: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/dev/lib/authority.js#L127

This means that to effectively change the domain of an existing credential provided by `TokenCredentialsBase`, we should provide a method to generate a new Authentication Context. From this, I'm proposing to add this method to that class:

```ts
  public setDomain(domain: string): void {
    this.domain = domain;
    const authorityUrl = this.environment.activeDirectoryEndpointUrl + this.domain;
    this._authContext = new AuthenticationContext(authorityUrl, this.environment.validateAuthority, this.tokenCache);
  }
```

Having that method, to facilitate new users to discover their tenants, this PR also proposes exposing the `buildTenantList` method.

A complete sample has also been provided where all of these changes come together to show how users can retrieve their subscriptions by any of their tenants in several ways, addressing the issue #89.

---

If approved:

Closes #89


lib/credentials/msiTokenCredentials.ts
```diff
@@ -116,7 +116,7 @@ export abstract class MSITokenCredentials implements TokenClientCredentials {
       if (typeof parsedBody["expires_on"] === "string") {
         // possibly a Date string '09/14/2017 00:00:00 PM +00:00'
         if (parsedBody["expires_on"].includes(":") || parsedBody["expires_on"].includes("/")) {
-          parsedBody.expiresOn = new Date(parsedBody["expires_on"], 10);
+          parsedBody.expiresOn = new Date(Number(parsedBody["expires_on"]), 10);
```
sadasant, line 119: I can't compile this project without this change.

samples/interactive-multiple-tenants.ts
```diff
@@ -0,0 +1,44 @@
...
+    const tenants = await msRestNodeAuth.buildTenantList(
+      initialAuthentication.credentials
+    );
+
+    // If you have already authenticated and you want to retrieve the subscriptions of a specific tenant,
+    // you can set the credential's domain to the tenant you want to use,
+    // then retrieve your tenant's subscriptions with the SubscriptionsClient from @azure/arm-subscriptions:
+    initialAuthentication.credentials.setDomain(tenants[0]);
+    const client = new SubscriptionClient(initialAuthentication.credentials);
+    const subscriptions = await client.subscriptions.list();
```
ramya-rao-a, line 28: What is the result of the list operation here when we don't call `setDomain()` on the credentials? Is an empty array returned or an error is thrown?

samples/interactive-multiple-tenants.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SubscriptionClient } from "@azure/arm-subscriptions";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+// copy the content of "sample.env" to a file named ".env". It should be stored at the root of the repo.
+// then from the root of the cloned repo run, ts-node ./samples/interactive.ts
+async function main(): Promise<void> {
+  try {
+    // Personal accounts will find no subscriptions on the initial authentication request.
+    // This is because the credentials are generated without specifying a tenant.
```
ramya-rao-a, line 11: It is more helpful if we can explain why credentials generated without specifying a tenant result in no subscriptions being listed. And why this is a problem with personal accounts only. I believe @schaabs mentioned something about v1 vs v2 apis and "organizations" tenant.

Related issues: https://github.com/Azure/azure-sdk-for-js/issues/7526 and https://github.com/Azure/azure-sdk-for-js/issues/4222

samples/interactive-multiple-tenants.ts
```diff
@@ -0,0 +1,44 @@
...
+    // you can set the credential's domain to the tenant you want to use,
+    // then retrieve your tenant's subscriptions with the SubscriptionsClient from @azure/arm-subscriptions:
+    initialAuthentication.credentials.setDomain(tenants[0]);
+    const client = new SubscriptionClient(initialAuthentication.credentials);
+    const subscriptions = await client.subscriptions.list();
+    console.log(`Subscriptions of tenant ${tenants[0]}`, subscriptions);
+
+    // You can also authenticate for a known tenant directly:
+    const tenantAuthentication = await msRestNodeAuth.interactiveLoginWithAuthResponse(
+      { domain: tenants[0] }
```
ramya-rao-a, line 33: If we want to showcase known tenants, then we shouldnt be using the value from the output of `buildTenantList`. 
I would keep the code commented, because this sample is meant for the case when the user does not know the tenant value

```js
// You can also authenticate for a known tenant directly by setting `domain` in the options
const knownTenantAuthResponse = await msRestNodeAuth.interactiveLoginWithAuthResponse({ domain: "<your-tenant-id>" });
```

samples/interactive-multiple-tenants.ts
```diff
@@ -0,0 +1,44 @@
+import * as msRestNodeAuth from "../lib/msRestNodeAuth";
+import { SubscriptionClient } from "@azure/arm-subscriptions";
+import * as dotenv from "dotenv";
+dotenv.config();
+
```
ramya-rao-a, line 5: The sample is named `interactive-multiple-tenants.ts`. Do we know for sure that the problem is when working with multiple tenants and not just personal accounts? For example, when you were able to replicate the issue, did you see multiple tenants?

lib/msRestNodeAuth.ts
```diff
@@ -11,7 +11,7 @@ export { MSIVmOptions, MSIVmTokenCredentials } from "./credentials/msiVmTokenCre
 export { TokenCredentialsBase } from "./credentials/tokenCredentialsBase";
 export { UserTokenCredentials } from "./credentials/userTokenCredentials";
 export { AuthConstants, TokenAudience } from "./util/authConstants";
-export { LinkedSubscription, LinkedUser, UserType } from "./subscriptionManagement/subscriptionUtils";
+export { LinkedSubscription, LinkedUser, UserType, buildTenantList } from "./subscriptionManagement/subscriptionUtils";
```
ramya-rao-a, line 14: Is `buildTenantList` the best name to use now that we are exposing this api?

ramya-rao-a:
@daviwil, @jonathandturner, Can you please review?
cc @schaabs

Changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 3.0.4 - 2020/05/19
+
+- Made `buildTenantList` public, to allow users to retrieve all of their tenants after authenticating.
+- Added a new public method to `TokenCredentialsBase` called `setDomain` that can be used to force Adal to renew the token against a different tenant.
```
ramya-rao-a, line 6: @daviwil, @jonathandturner, @schaabs Does this feature of changing the tenant to be used in a credential have a place in the new world of Track 2 sdks that we are working on?

Same question for the `buildTenantList`. If we are saying that customers need to know the tenant id before doing auth, are we willing to expose a helper method to do so in Track 2?

samples/interactive-multiple-tenants.ts
sadasant: The problem is when working with personal accounts. My intention with the name of the file was to showcase how to authenticate with multiple tenants in multiple ways. It seems to me that it's kind of weird to have a sample just to showcase that a bug is fixed. Perhaps it's more valuable to make it more generic? In this case: to show how to do multiple things with multiple tenants?

lib/msRestNodeAuth.ts
sadasant: What about `getTenants()`?

samples/interactive-multiple-tenants.ts
sadasant: If we don't call `setDomain()`, the result of the list operation is `[ nextLink: undefined ]`. No error is thrown.

samples/interactive-multiple-tenants.ts
sadasant: I think this is correct, but please help me, @schaabs 

> Personal accounts will find no subscriptions on the initial authentication request.
> This is because the credentials are generated without specifying a tenant.
> 
> For more context:
> When we authenticate with the Azure Active Directory (AAD), it attempts to associate
> the authenticating user with a target "directory". This directory is specified by the given "domain".
> If no domain is specified, the "common" domain is assumed.
> In this common domain, while organizations might have shared resources, personal accounts will show nothing.
> In AAD v2, the "organizations" domain was added, to allow users to authenticate and view all of the available resources
> regardless of the type of account they might have.
> Since ms-rest-nodeauth only supports AAD v1, we have to change the domain after authenticating, as shown in this sample.
> 
> Our new `@azure/identity` package provides support for AAD v2:
> https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/identity/identity

Once we agree on an approach, I can update this PR.

Changelog.md
daviwil: `tenantId` is a core part of the AAD credential APIs in `@azure/identity`, for example: https://azuresdkdocs.blob.core.windows.net/$web/javascript/azure-identity/1.0.0/classes/clientsecretcredential.html#constructor.  Because a developer needs an App Registration to set up AAD auth in their application, they will have a tenand ID because one is provided to them with their App Registration.

lib/credentials/tokenCredentialsBase.ts
```diff
@@ -8,7 +8,7 @@ import { TokenClientCredentials } from "./tokenClientCredentials";
 import { TokenResponse, AuthenticationContext, MemoryCache, ErrorResponse, TokenCache } from "adal-node";
 
 export abstract class TokenCredentialsBase implements TokenClientCredentials {
-  public readonly authContext: AuthenticationContext;
```
daviwil, line 11: Changing this to `private` is an API-breaking change, maybe we shouldn't do it?

samples/interactive-multiple-tenants.ts
daviwil: The text you wrote sounds accurate enough to me, but it may be more accurate to say:

> Personal accounts will return no subscriptions on the initial authentication request unless `domain` is specified.

ramya-rao-a: But is "personal" account an official description? For example, in https://github.com/Azure/azure-sdk-for-js/issues/7526 there is reference to MSA accounts. In https://github.com/Azure/azure-sdk-for-js/issues/4222, there is reference to "live" accounts. How are these accounts officially documented?

daviwil: I think "Live ID" is referring to personal Microsoft accounts, but "personal" account refers more broadly to what they call "B2C" account which might be federated user accounts like Google, Facebook, etc.  The distinction between different account types is here: 

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-overview

For this particular case it may be more accurate to say "Personal Microsoft accounts" since that's the more likely scenario.  It is possible to have B2C accounts assigned to a tenant but you need to create a special tenant for that:

https://docs.microsoft.com/en-us/azure/active-directory-b2c/tutorial-create-tenant

lib/credentials/tokenCredentialsBase.ts
ramya-rao-a: Actually it is getting replaced with a getter by the same name. Would that still be breaking change?

daviwil: Ahhh...  in that case, I'm not sure.  @bterlson, do you know if changing a public field to a public property would be considered a breaking change in TypeScript?

bterlson: Generally it's not breaking to move from a regular property to an accessor. In rare circumstances it could be, e.g. if users had a reason to grab the descriptor via Object.getOwnPropertyDescriptor and process it, but that doesn't apply here. But from a typescript perspective, this shouldn't break anyone's code since `readonly` was present before.

Worth noting that this is a breaking change from the perspective of a JS developer, as previously they could set `authContext` without getting an error, and now that will throw. I would guess that setting it had no effect before?

samples/interactive-multiple-tenants.ts
ramya-rao-a: > My intention with the name of the file was to showcase how to authenticate with multiple tenants in multiple ways.

What are the multiple ways being referred to here? I see only one way, which is to 
- create credential
- fetch tenants using the credential
- call setDomain() to update the tenant used by the credential
- then use the credential with Azure SDKs

sadasant: The multiple ways come from this, which has been now commented:

```
// You can skip all of the above, if you already know the tenant id
// const tenantAuthentication = await msRestNodeAuth.interactiveLoginWithAuthResponse({ domain: "<your-tenant-id>" });
```

Would a name closer to what you're expecting be: `interactive-manual-tenant`?

lib/credentials/tokenCredentialsBase.ts
sadasant: I'll undo this change. This was an opinionated change by which I wanted to highlight that allowing users to change the authContext was perhaps a misdirection, since they could end up with many buggy or broken results. I'll make this a public property again.

sadasant: Setting authContext could have many behaviors. I'm guessing that someone very determined could make something weird enough to be considered useful with this. Though I have to admit that this change was mostly opinionated. The alternative is to make it a public property and remove the `readonly` part (so I can set this property again.

samples/interactive-multiple-tenants.ts
ramya-rao-a: From where I stand, we need to communicate the below:
- The `subscriptions` in the AuthResponse will be an empty list for personal accounts.
- In case of personal accounts credentials created without passing a tenantId don't work as expected when passed to other Azure SDKs.
- People can get the tenant id from the Azure portal or Azure CI (not sure, I havent tried). 
- But, if one needs to get their tenantId programatically, they can now use the new `buildTenantsList` on the credential to get the list, pick the one they want and then re-create their credentials with the tenantId or update the existing credentials using `setDomain()`.

Now, its not clear to me whether having a sample is the right way to communicate all of the above. The readme should be good enough as well, it would get more visibility.

ramya-rao-a: In addition to readme, we can also update the doc strings for some of the above
- `subscriptions` in `AuthResponse`
- The tenantId/domain argument in input arguments for each of the login methods

lib/msRestNodeAuth.ts
ramya-rao-a: Now that we accidentally shipped 3.0.4, we can stick with `buildTenantList` so as to not break anyone already on 3.0.4

Changelog.md
ramya-rao-a: So if one didnt know the tenantId, they could pass `organizations` or something like that as tenantId in `@azure/identity` and things would magically work because we are using v2 apis of AAD?

lib/credentials/tokenCredentialsBase.ts
ramya-rao-a: How about instead of `setDomain()` changing the auth context, we return a new credentials with the right tenantId and so the right auth context? The method would need a new name though

Changelog.md
daviwil: Yep, they should be able to pass `organizations` there.  However, we don't have the same subscription-listing behavior that `ms-rest-nodeauth` does because it was never a part of the design, so we wouldn't have hit this issue anyway.

lib/msRestNodeAuth.ts
daviwil: Ouch...  Since this isn't super likely to have been used yet I'd be willing to change it anyway, but I can understand not doing it since we technically shipped it.

lib/credentials/tokenCredentialsBase.ts
bterlson: I think generally speaking the credential types encapsulate the changing things inside of them (e.g. key credential rolling => the .key property changes, rather than the whole credential gets replaced).

I think just removing `readonly` is the safest change to make here. If there is strong reason to move to a getter (e.g. because setting the property does nothing or interferes in some bad way) then I could be convinced its fine, but if it's mostly being conservative/defensive, that ship has sailed already.

Changelog.md
ramya-rao-a: The subscription listing in the response was just a part of the problem. The other part was that the token generated using the default tenantId in ms-rest-nodeauth just wouldnt work. So, for such cases in `azure/identity`, we can ask people to pass `organizations`.

ramya-rao-a: > The subscription listing in the response was just a part of the problem. The other part was that the token generated using the default tenantId in ms-rest-nodeauth just wouldnt work.

@sadasant, Please correct me if I got the above wrong

ramya-rao-a:
The build is probably failing as we need to increment the package version again

Changelog.md
sadasant: The token generated by the default subscription didn't work because of the domain used. "common" doesn't show the user resources properly on personal accounts, at least in the previous version of the OAuth service provider. To correct this behavior in track 2, we need to take advantage of the new domain "organizations". @davidwil do you know what domain we're using by default in track 2?

lib/credentials/tokenCredentialsBase.ts
sadasant: sounds good then. I'll remove `readonly`

sadasant: Removed `readonly` here: https://github.com/Azure/ms-rest-nodeauth/pull/91/commits/4c39368a28f924b346fd067e60c0bce17078ccfd

samples/interactive-multiple-tenants.ts
sadasant: I think I addressed it here: https://github.com/Azure/ms-rest-nodeauth/pull/91/commits/8254af1f70f9683b5e8460bcd04b19229112ce98 but please let me know if I missed something!

Changelog.md
ramya-rao-a: In Track 2, tenantId is mandatory

daviwil: In most cases we don't use one by default in `@azure/identity`, we always ask the user to provide it.  There's one exception though -- the `InteractiveBrowserCredential`, which is effectively the same scenario as the bug you're trying to fix. Lo and behold, we're also using `common` as the default in that case:

https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/identity/identity/src/constants.ts#L19

Apologies for not remembering that.  We need to verify with @schaabs whether this should actually be `organizations` by default.

sadasant: I can test this and make a PR. In fact, I will do that now 🌞

README.md
```diff
@@ -1,8 +1,16 @@
...
 
 This library provides different node.js based authentication mechanisms for services in Azure. It also contains rich type definitions thereby providing a good TypeScript experience.
+
 All the authentication methods support callbacks as well as promises. If they are called within an async method in your application then you can use the async/await pattern as well.
 
+Things to consider:
+
+- Many of these authentication methods accept a `domain` to be specified, which accepts `tenants` to be passed in.
+- For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically.
+- Users need to call to the `setDomain` method of the returned credential with the tenant of their preference in order to access their resources.
```
daviwil, line 11:
```suggestion
- In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential.  Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```

README.md
```diff
@@ -1,8 +1,16 @@
 # ms-rest-nodeauth [![Build Status](https://dev.azure.com/azure-public/adx/_apis/build/status/public.Azure.ms-rest-nodeauth)](https://dev.azure.com/azure-public/adx/_build/latest?definitionId=9)
 
 This library provides different node.js based authentication mechanisms for services in Azure. It also contains rich type definitions thereby providing a good TypeScript experience.
+
 All the authentication methods support callbacks as well as promises. If they are called within an async method in your application then you can use the async/await pattern as well.
 
+Things to consider:
+
+- Many of these authentication methods accept a `domain` to be specified, which accepts `tenants` to be passed in.
+- For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically.
```
daviwil, line 10:
```suggestion
- When authenticating a user with a personal Microsoft account without specifying a `domain` in the credential, the received token won't be able to see resources belonging to subscriptions in your Azure tenant.
```

README.md
```diff
@@ -1,8 +1,16 @@
...
+- Users need to call to the `setDomain` method of the returned credential with the tenant of their preference in order to access their resources.
+- Credentials can be sent to the `buildTenantList` method to obtain the list of available tenants for any account.
```
daviwil, line 12: I'd remove this line entirely if you accept the above two suggestions.

README.md
sadasant: @daviwil before I add this commit, I have more or less the same wording in the code comments, should I do this change there too?

daviwil: Yep, better to be consistent.

sadasant: I pushed an update, but I will review again in the morning.

lib/credentials/tokenCredentialsBase.ts
ramya-rao-a: https://github.com/Azure/ms-rest-nodeauth/commit/4c39368a28f924b346fd067e60c0bce17078ccfd is doing more than just removing readonly. We are now allowing `authContext` to be undefined, which then is requiring us to add null/undefined checks and throw errors. This seems unnecessary.

Is this because `authContext` is not being set in the constructor directly anymore as we call `setDomain()` instead? If so, then what was the motivation for this change? If the answer is to re-use the method `setDomain()`, then I would argue that the downside of this (the if checks for missing authcontexts, which would never be true for the real world user) outweighs the benefits of re-using a method.

Changelog.md
sadasant: @daviwil I have been trying to use the interactive browser authentication, but I haven't been able to. Not sure how we're distributing these files if our package.jsons don't have the dist-browser folder in them, and if I export my files and try to use it from a website, when identity goes to do the browser roundtrip, it fails with a very strange error on the Azure login page. I think I need this to show this to you.

Long story short, if you know an easy way to test it, you can probably do that faster than me.

daviwil:
Here's another issue reported from the accidental release of the contents of this PR.  @sadasant can you double check and make sure that the latest changes here resolve the issue they saw with 3.0.4?

https://github.com/Azure/ms-rest-nodeauth/issues/94

lib/credentials/tokenCredentialsBase.ts
ramya-rao-a: > now that this property is public, users can simply set it to undefined and crash their own experience 

That can be said for every public property on every class on every package we have. We don't do checks for undefined/null for public properties unless we mark the type to allow for undefined. 

In a JS program, a user can update a property even if it was marked as readonly. So, even before this PR, a JS user could set authContext to undefined. Setting properties on a class to invalid values resulting in the program behaving unexpectedly is fine. Especially if the error has a good enough stack trace to help the user figure out what went wrong. I don't see why we should change the status quo here.

sadasant: I am personally in favor of doing these type checks in our new projects too. Though at least I think these deprecated projects could receive some extra help to avoid having unintended consequences.

lib/credentials/msiTokenCredentials.ts
sadasant: I'll change it to `parseInt` which is used by some lines below.

sadasant:
With the recent updates, this PR also fixes:

Fixes #88
by adding more documentation to the MSI samples in the README.

Fixes #86
by the change introduced here https://github.com/Azure/ms-rest-nodeauth/pull/85

lib/credentials/tokenCredentialsBase.ts
sadasant: -- I will duplicate the code and remove the type validation

sadasant:
@daviwil I can't reproduce that issue with this PR.

lib/login.ts
```diff
@@ -394,14 +406,18 @@ export async function withAuthFileWithAuthResponse(options?: LoginWithAuthFileOp
  * Provides a url and code that needs to be copy and pasted in a browser and authenticated over there. If successful, the user will get a
  * DeviceTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
  *
+ * The `subscriptions` property in the AuthResponse will be an empty array for personal accounts.
+ *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
+ *
```
ramya-rao-a, line 412:
```suggestion
 * DeviceTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants. If no domain i.e tenantId is passed in the options when working with personal accounts, 
the list of subscriptions will be empty.
 *
```

README.md
```diff
@@ -1,8 +1,16 @@
...
 This library provides different node.js based authentication mechanisms for services in Azure. It also contains rich type definitions thereby providing a good TypeScript experience.
+
 All the authentication methods support callbacks as well as promises. If they are called within an async method in your application then you can use the async/await pattern as well.
 
+Things to consider:
+
+- Many of these authentication methods accept a `domain` to be specified, which accepts `tenants` to be passed in.
+- For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically.
+- In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential.
+  Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```
ramya-rao-a, line 12:
```suggestion
Things to consider when using personal accounts:

The authentication methods accept a `domain` in the options parameter where you can pass the id of your tenant. When using personal accounts, credentials created with no `domain` fail to generate the right token for authentication. Due to the same reason, the list of subscriptions for all the tenants in your account returned by some of these methods will be empty too.

The workaround is to pass the tenant Id to `domain` property in the options when using these authentication methods. You can get the tenant id from Azure portal or Azure CLI.

If you need to fetch the tenantId programatically, follow the below steps
- Use any of the authentication methods without setting the domain to get a credential
- Pass the credential the `buildTenantLists()` to get the list of all tenants in your account

You can also update the `domain` in the same credential by using the `setDomain()` method on it.
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,51 @@
...
+    // If no domain is specified, the "common" domain is assumed.
+    // In this common domain, while organizations might have shared resources, personal accounts will show nothing.
+    // In AAD v2, the "organizations" domain was added, to allow users to authenticate and view all of the available resources
+    // regardless of the type of account they might have.
+    // Since ms-rest-nodeauth only supports AAD v1, we have to change the domain after authenticating, as shown in this sample.
+    //
+    // Our new `@azure/identity` package provides support for AAD v2:
+    // https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/identity/identity
+
+    const initialAuthentication = await msRestNodeAuth.interactiveLoginWithAuthResponse();
```
ramya-rao-a, line 25: We don't call interactive login again after the initial call, so we shouldnt need to use the word "initial" in the name here. How about just `authResponse`?

lib/login.ts
```diff
@@ -112,7 +112,7 @@ export interface AuthResponse {
    */
   credentials: TokenCredentialsBase;
   /**
-   * @property {Array<LinkedSubscription>} [subscriptions] List of associated subscriptions.
+   * @property {Array<LinkedSubscription>} [subscriptions] List of associated subscriptions. Caution: This list may be empty if the credential was created without a domain i.e tenantId when using personal accounts.
```
ramya-rao-a, line 115: I see now that this can be confusing as one might try to see where to find the `tenantId` property.

```suggestion
   * @property {Array<LinkedSubscription>} [subscriptions] List of associated subscriptions. Caution: This list may be empty if the credential was created without setting the domain to the tenant id when using personal accounts.
```

lib/login.ts
```diff
@@ -144,19 +144,21 @@ export type Callback<TResult> = (error?: Error, result?: TResult) => void;
  * Provides a UserTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
  * This method is applicable only for organizational ids that are not 2FA enabled otherwise please use interactive login.
  *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
+ *
```
ramya-rao-a, line 148: As mentioned in https://github.com/Azure/ms-rest-nodeauth/pull/91#discussion_r437629333, the readme talks about how to use buildTenantList, we do not need to have that information in docs for every auth method.

```suggestion
 * Provides a UserTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants. 
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
 * This method is applicable only for organizational ids that are not 2FA enabled otherwise please use interactive login.
 *
```

lib/login.ts
```diff
@@ -144,19 +144,21 @@ export type Callback<TResult> = (error?: Error, result?: TResult) => void;
...
- * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferrably in a guid format).
+ * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferably in a guid format).
  * @param {string} [options.domain] The domain or tenant id containing this application. Default value "common".
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains `credentials` and an optional `subscriptions` array, and rejects with an Error. If a personal account is used to authenticate, `subscriptions` will return empty.
```
ramya-rao-a, line 161: As mentioned in https://github.com/Azure/ms-rest-nodeauth/pull/91#discussion_r437645521, we need to mention that this happens only when domain is not set.

```suggestion
 * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains `credentials` and an optional `subscriptions` array, and rejects with an Error. The `subscriptions` array in the AuthResponse will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -175,15 +177,17 @@ export async function withUsernamePasswordWithAuthResponse(username: string, pas
...
   const subscriptionList: LinkedSubscription[] = await _getSubscriptions(creds, tenantList, options.tokenAudience);
 
   return { credentials: creds, subscriptions: subscriptionList };
 }
 
 /**
- * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrinicpalId/clientId across all the applicable tenants.
+ * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
+ *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```
ramya-rao-a, line 190:
```suggestion
 * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants. 
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
 *
```

lib/login.ts
```diff
@@ -192,11 +196,11 @@ export async function withUsernamePasswordWithAuthResponse(username: string, pas
...
  * @param {string} [options.tokenAudience] The audience for which the token is requested. Valid values are 'graph', 'batch', or any other resource like 'https://vault.azure.net/'.
- * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferrably in a guid format).
+ * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferably in a guid format).
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. If a personal account is used to authenticate, `subscriptions` will return empty.
```
ramya-rao-a, line 203:
```suggestion
 * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. The `subscriptions` array in the AuthResponse will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -215,7 +219,9 @@ export async function withServicePrincipalSecretWithAuthResponse(clientId: strin
 }
 
 /**
- * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrinicpalId/clientId across all the applicable tenants.
+ * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
+ *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```
ramya-rao-a, line 224:
```suggestion
 * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
```

lib/login.ts
```diff
@@ -226,11 +232,11 @@ export async function withServicePrincipalSecretWithAuthResponse(clientId: strin
...
  * @param {string} [options.tokenAudience] The audience for which the token is requested. Valid values are 'graph', 'batch', or any other resource like 'https://vault.azure.net/'.
- * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferrably in a guid format).
+ * If tokenAudience is 'graph' then domain should also be provided and its value should not be the default 'common' tenant. It must be a string (preferably in a guid format).
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. If a personal account is used to authenticate, `subscriptions` will return empty.
```
ramya-rao-a, line 239:
```suggestion
 * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. The `subscriptions` array in the AuthResponse will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -416,7 +423,7 @@ export async function withAuthFileWithAuthResponse(options?: LoginWithAuthFileOp
  *
  * @param {function} [optionalCallback] The optional callback.
  *
- * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse that contains "credentials" and optional "subscriptions" array and rejects with an Error.
+ * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. The "subscriptions" array in the AuthResponse will be empty if no domain is passed in the options when using with personal accounts.
```
ramya-rao-a, line 426:
```suggestion
 * @returns {Promise<AuthResponse>} A Promise that resolves to AuthResponse, which contains "credentials" and optional "subscriptions" array and rejects with an Error. The `subscriptions` array in the AuthResponse will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -575,12 +582,14 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
  * Provides a url and code that needs to be copy and pasted in a browser and authenticated over there. If successful, the user will get a
  * DeviceTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
  *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
+ *
```
ramya-rao-a, line 586:
```suggestion
 * Provides a url and code that needs to be copy and pasted in a browser and authenticated over there. If successful, the user will get a
 * DeviceTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
 *
```

lib/login.ts
```diff
@@ -624,7 +633,9 @@ export function interactive(options?: InteractiveLoginOptions, callback?: { (err
 }
 
 /**
- * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrinicpalId/clientId across all the applicable tenants.
+ * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
+ *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```
ramya-rao-a, line 638:
```suggestion
 * Provides an ApplicationTokenCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
```

lib/login.ts
```diff
@@ -592,9 +601,9 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
  * @returns {function | Promise} If a callback was passed as the last parameter then it returns the callback else returns a Promise.
  *
  *    {function} optionalCallback(err, credentials)
- *                 {Error}  [err]                           - The Error object if an error occurred, null otherwise.
- *                 {DeviceTokenCredentials} [credentials]   - The DeviceTokenCredentials object.
- *                 {Array}                [subscriptions]   - List of associated subscriptions across all the applicable tenants.
+ *                 {Error}                          [err]  - The Error object if an error occurred, null otherwise.
+ *                 {DeviceTokenCredentials} [credentials]  - The DeviceTokenCredentials object.
+ *                 {Array}                [subscriptions]  - List of associated subscriptions across all the applicable tenants. If a personal account is used to authenticate, this array will return empty.
```
ramya-rao-a, line 606:
```suggestion
 *                 {Array}                [subscriptions]  - List of associated subscriptions across all the applicable tenants. This list will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -592,9 +601,9 @@ export function withAuthFile(options?: LoginWithAuthFileOptions, callback?: { (e
  * @returns {function | Promise} If a callback was passed as the last parameter then it returns the callback else returns a Promise.
  *
  *    {function} optionalCallback(err, credentials)
- *                 {Error}  [err]                           - The Error object if an error occurred, null otherwise.
- *                 {DeviceTokenCredentials} [credentials]   - The DeviceTokenCredentials object.
- *                 {Array}                [subscriptions]   - List of associated subscriptions across all the applicable tenants.
+ *                 {Error}                          [err]  - The Error object if an error occurred, null otherwise.
+ *                 {DeviceTokenCredentials} [credentials]  - The DeviceTokenCredentials object.
```
ramya-rao-a, line 605:
```suggestion
 *                 {DeviceTokenCredentials} [credentials]  - The DeviceTokenCredentials object. This may generate invalid tokens if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -633,17 +644,17 @@ export function interactive(options?: InteractiveLoginOptions, callback?: { (err
...
  * @param {Environment} [options.environment] The azure environment to authenticate with.
  * @param {object} [options.tokenCache] The token cache. Default value is the MemoryCache object from adal.
  * @param {function} [optionalCallback] The optional callback.
  *
  * @returns {function | Promise} If a callback was passed as the last parameter then it returns the callback else returns a Promise.
  *
  *    {function} optionalCallback(err, credentials)
- *                 {Error}  [err]                               - The Error object if an error occurred, null otherwise.
+ *                 {Error}                               [err]  - The Error object if an error occurred, null otherwise.
  *                 {ApplicationTokenCredentials} [credentials]  - The ApplicationTokenCredentials object.
```
ramya-rao-a, line 656:
```suggestion
 *                 {ApplicationTokenCredentials} [credentials]  - The ApplicationTokenCredentials object. This may generate invalid tokens if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -633,17 +644,17 @@ export function interactive(options?: InteractiveLoginOptions, callback?: { (err
...
- *                 {Array}                [subscriptions]       - List of associated subscriptions across all the applicable tenants.
+ *                 {Array}                     [subscriptions]  - List of associated subscriptions across all the applicable tenants. If a personal account is used to authenticate, this array will return empty.
```
ramya-rao-a, line 657:
```suggestion
 *                 {Array}                     [subscriptions]  - List of associated subscriptions across all the applicable tenants. This list will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -673,7 +684,9 @@ export function withServicePrincipalSecret(clientId: string, secret: string, dom
 }
 
 /**
- * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrinicpalId/clientId across all the applicable tenants.
+ * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
+ *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
```
ramya-rao-a, line 689:
```suggestion
 * Provides an ApplicationTokenCertificateCredentials object and the list of subscriptions associated with that servicePrincipalId/clientId across all the applicable tenants.
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
```

lib/login.ts
```diff
@@ -694,7 +707,7 @@ export function withServicePrincipalSecret(clientId: string, secret: string, dom
  *    {function} optionalCallback(err, credentials)
  *                 {Error}  [err]                               - The Error object if an error occurred, null otherwise.
  *                 {ApplicationTokenCertificateCredentials} [credentials]  - The ApplicationTokenCertificateCredentials object.
- *                 {Array}                [subscriptions]       - List of associated subscriptions across all the applicable tenants.
+ *                 {Array}                [subscriptions]       - List of associated subscriptions across all the applicable tenants. If a personal account is used to authenticate, this array will return empty.
```
ramya-rao-a, line 710:
```suggestion
 *                 {Array}                [subscriptions]       - List of associated subscriptions across all the applicable tenants. This list will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -694,7 +707,7 @@ export function withServicePrincipalSecret(clientId: string, secret: string, dom
  *    {function} optionalCallback(err, credentials)
  *                 {Error}  [err]                               - The Error object if an error occurred, null otherwise.
  *                 {ApplicationTokenCertificateCredentials} [credentials]  - The ApplicationTokenCertificateCredentials object.
```
ramya-rao-a, line 709:
```suggestion
 *                 {ApplicationTokenCertificateCredentials} [credentials]  - The ApplicationTokenCertificateCredentials object. This may generate invalid tokens if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -727,14 +740,16 @@ export function withServicePrincipalCertificate(clientId: string, certificateStr
  * Provides a UserTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
  * This method is applicable only for organizational ids that are not 2FA enabled otherwise please use interactive login.
  *
+ * For personal accounts, credentials created without passing a domain won't be able to access most of the account resources automatically. In this case you will need to call `buildTenantList` to gather the list of tenants so that the ID of one of them can be passed into the `setDomain` method of the returned credential. Once the domain is set, you will be able to access resources from subscriptions in that tenant.
+ *
```
ramya-rao-a, line 744:
```suggestion
 * Provides a UserTokenCredentials object and the list of subscriptions associated with that userId across all the applicable tenants.
 * This list may be empty and the credentials may generate invalid tokens when using personal accounts if `domain` is not set to the tenant id in the `options`.
 * This method is applicable only for organizational ids that are not 2FA enabled otherwise please use interactive login.
 *
```

lib/login.ts
```diff
@@ -745,7 +760,7 @@ export function withServicePrincipalCertificate(clientId: string, certificateStr
  *    {function} optionalCallback(err, credentials)
  *                 {Error}  [err]                         - The Error object if an error occurred, null otherwise.
  *                 {UserTokenCredentials} [credentials]   - The UserTokenCredentials object.
- *                 {Array}                [subscriptions] - List of associated subscriptions across all the applicable tenants.
+ *                 {Array}                [subscriptions] - List of associated subscriptions across all the applicable tenants. If a personal account is used to authenticate, this array will return empty.
```
ramya-rao-a, line 763:
```suggestion
 *                 {Array}                [subscriptions] - List of associated subscriptions across all the applicable tenants. This list will be empty if no domain is passed in the options when using with personal accounts.
```

lib/login.ts
```diff
@@ -745,7 +760,7 @@ export function withServicePrincipalCertificate(clientId: string, certificateStr
  *    {function} optionalCallback(err, credentials)
  *                 {Error}  [err]                         - The Error object if an error occurred, null otherwise.
  *                 {UserTokenCredentials} [credentials]   - The UserTokenCredentials object.
```
ramya-rao-a, line 762:
```suggestion
 *                 {UserTokenCredentials} [credentials]   - The UserTokenCredentials object. This may generate invalid tokens if no domain is passed in the options when using with personal accounts
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as dotenv from "dotenv";
+dotenv.config();
+
+// copy the content of "sample.env" to a file named ".env". It should be stored at the root of the repo.
+// then from the root of the cloned repo run, ts-node ./samples/interactive.ts
+
+async function main(): Promise<void> {
+  try {
+    // Personal accounts will return no subscriptions on the initial authentication request unless domain is specified.
+    // This is because the credentials are generated without specifying a tenant.
```
ramya-rao-a, line 22: We don't need this here as we have the comments in the starting of the sample that provides the same information.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,65 @@
...
+    //
+    // For more context:
```
ramya-rao-a, line 24: For more context, we can point to the issue that was logged and make sure that issue has all the information needed.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,65 @@
...
+    // When we authenticate with the Azure Active Directory (AAD), it attempts to associate
+    // the authenticating user with a target "directory". This directory is specified by the given "domain".
+    // If no domain is specified, the "common" domain is assumed.
+    // In this `common` domain, while organizations might have shared resources, personal accounts will show nothing.
+    // In AAD v2, the "organizations" domain was added, to allow users to authenticate and view all of the available resources
+    // regardless of the type of account they might have.
+    // Since `ms-rest-nodeauth` only supports AAD v1, we have to change the domain after authenticating, as shown in this sample file.
+    //
+    // Our new `@azure/identity` package provides support for AAD v2:
```
ramya-rao-a, line 33: I wouldnt talk about `@azure/identity` here as none of the users of `ms-rest-nodeauth` can use the newer `@azure/identity` package.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,65 @@
...
+    // https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/identity/identity
+    const authentication = await msRestNodeAuth.interactiveLoginWithAuthResponse();
+    console.log(
```
ramya-rao-a, line 36:
```suggestion
    // For personal accounts, the below will print an empty array as we did not set the domain
    console.log(
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,65 @@
...
+      "Subscriptions retrieved by default",
+      authentication.subscriptions
+    );
+
+    // For personal accounts, the following code will return empty subscriptions:
```
ramya-rao-a, line 41: Based on our previous conversations, my understanding was that we would get an error regarding invalid tokens when using credentials created without setting the domain. Is that not the case? If so, then half of my suggested edits in login.ts file are not applicable.

ramya-rao-a: After offline conversation, we found that `client.subscriptions.list()` does not throw error, but `client.subscriptions.get()` does throw an error.


We cant use the get() because we dont have the subscription id here. We can do one of two things
- Use some other client that throws the expected error (I doubt we can find it, almost all clients need the subscription id at the bare minimum)
- Have the user fill in the subscription id in the sample. They can find it in the Azure portal

lib/login.ts
sadasant: I believe that I didn't see the original suggestion from a web cache issue. I usually try to follow up, even with a comment or a reaction.

I don't think this is correct `may generate invalid tokens` though. I'll post my proposal in a bit.

samples/interactivePersonalAccount.ts
sadasant: Does this mean removing this text?

sadasant: I added the text below here: https://github.com/Azure/ms-rest-nodeauth/issues/89 I will remove this text then.

samples/interactivePersonalAccount.ts
sadasant: Filling-in the subscription in the sample works for me.

Changelog.md
```diff
@@ -1,5 +1,16 @@
...
 
+## 3.0.5 - (unreleased)
+
+- Fixed breaking changes introduced in 3.0.4.
+- Fixed typos in error messages.
+
+## 3.0.4 - 2020/05/19 (deprecated)
+
+- Made `buildTenantList` public, to allow users to retrieve all of their tenants after authenticating.
+- Added a new public method to `TokenCredentialsBase` called `setDomain` that can be used to force Adal to renew the token against a different tenant.
```
ramya-rao-a, line 11: Let's move this to 3.0.5
Changelog is for people to know why they should update to a specific version. Since 3.0.4 is marked as deprecated, it is not clear if the changes made there will be available in 3.0.5

Under 3.0.4, we can add a note on the accidental release that had changes related to work in progress.
The note on sample is not needed as it is not part of the package.

Under 3.0.5, we don't need to specify the breaking changes as there is none compared to 3.0.3.

lib/login.ts
sadasant: We agreed on not using this line through teams.

lib/login.ts
sadasant: We agreed on not using this line through teams.

Changelog.md
sadasant: I think I did this! Based on this feedback, of course.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,69 @@
...
+  console.log(
+    "Subscriptions retrieved by default",
+    authentication.subscriptions
+  );
+
+  const client = new SubscriptionClient(authentication.credentials);
+
+  // For personal accounts, this will return an empty array:
+  let subscriptions = await client.subscriptions.list();
+  console.log(`These subscriptions will be empty for personal accounts`, subscriptions);
```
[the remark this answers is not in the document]
sadasant: On it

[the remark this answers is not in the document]
sadasant: I believe I did what you meant. Let me know if I should do something else.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+ * If you need to fetch the tenant Id programmatically:
+ *
+ * - Use any of the authentication methods without setting the domain to get a credential.
+ * - Call the `buildTenantLists(credential)` method by sending that same credential as the first parameter to get the list of all tenants in your account.
+ *
+ * If you want to avoid having to create a new credential, once you have access to one of the account's tenants, you can set it as the `domain` of your credentials by using the method `setDomain(tenant)` from the existing credentials object.
+ *
+ * This sample shows when the authentication fails for personal accounts, and how to overcome this issue.
+ *
+ * For context: https://github.com/Azure/ms-rest-nodeauth/issues/89#issuecomment-643471343
```
ramya-rao-a, line 15: Updating the text here to match what is in the readme and removing the steps on how to use buildTenantList as that is shown in the code

```suggestion
 * When using personal accounts, the `domain` property in the options passed to the authentication methods is mandatory and should be set to the tenant id. If this property is not set, the credentials created by the authentication methods will not be able to access any of the resources of the personal account. For that same reason, the list of subscriptions expected in the return value of these methods will be empty.
 * 
 * You can get the tenant Id from Azure portal or the Azure CLI. If you need to fetch the tenant Id programmatically, you can use the helper method `buildTenantList(credential)`.
 *
 * This sample shows how the authentication fails for personal accounts when `domain` is not set, and how to overcome this issue.
 *
 * For more context: https://github.com/Azure/ms-rest-nodeauth/issues/89#issuecomment-643471343
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+ */
+import * as msRestNodeAuth from "../lib/msRestNodeAuth";
+import { SubscriptionClient } from "@azure/arm-subscriptions";
+import * as dotenv from "dotenv";
+dotenv.config();
+
+// Copy the content of "sample.env" to a file named ".env". It should be stored at the root of the repo.
+// then from the root of the cloned repo run, ts-node ./samples/interactive.ts
```
ramya-rao-a, line 23: Does this sample depend on any environment variable? If not, I would suggest removing the above comment and the importing of `dotenv`.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+
+async function main(): Promise<void> {
+  const authentication = await msRestNodeAuth.interactiveLoginWithAuthResponse();
```
ramya-rao-a, line 26: All other samples are using the name `authres` for the output. Can we use the same? Another options is `authResponse`.

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+  console.log(
+    "Subscriptions retrieved by default",
+    authentication.subscriptions
+  );
+
+  const client = new SubscriptionClient(authentication.credentials);
+
+  // Note: You must replace `<my-subscription>` with the Id of one of your subscriptions to get the error.
+  // If you use this code as it is, you will get `The subscription '<my-subscription>' could not be found.`,
+  // which is not an error relevant to the issue we're describing.
```
ramya-rao-a, line 38: We shouldnt need to go into these details. The error message `The subscription '<my-subscription>' could not be found.` is self explanatory

```suggestion
  // Note: Replace `<my-subscription>` below with the Id of one of your subscriptions
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+  const subscriptionId = "<my-subscription>";
+
+  // Some requests will fail for personal accounts until we update the domain on the credential.
```
ramya-rao-a, line 41:
```suggestion
  // The below request to get the subscription details will fail for personal accounts until we update the domain on the credential.
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  // To get the tenants for your account, you can use the buildTenantList method:
+  const tenants = await msRestNodeAuth.buildTenantList(
+    authentication.credentials
+  );
+
+  // If you have already authenticated and you want to retrieve the subscriptions of a specific tenant,
+  // you can set the credential's domain to the tenant you want to use,
+  // then retrieve your tenant's subscriptions with the SubscriptionsClient from @azure/arm-subscriptions:
```
ramya-rao-a, line 59: By this point, the user has already authenticated, so we shouldnt need to talk about "If you have already authenticated..."

```suggestion
  // Update the domain used by the credentials so that it can generate token against the right tenant
```

samples/interactivePersonalAccount.ts
```diff
@@ -0,0 +1,74 @@
...
+  const subscription = await client.subscriptions.get(subscriptionId);
+  console.log("After specifying the tenant, we're able to retrieve the full information of our subscriptions:", subscription);
+
+  // You can skip all of the above if you already know the tenant Id, and do something like the following:
+  const tenantAuthentication = await msRestNodeAuth.interactiveLoginWithAuthResponse({ domain: "<your-tenant-id>" });
+  console.log(
+    "Subscriptions retrieved after authenticating with a specific domain",
+    tenantAuthentication.subscriptions
+  );
+}
```
ramya-rao-a, line 72: I think we can skip this. We began saying that one needs to set the domain to the tenant id, so we don't need to demonstrate it here.

```suggestion
```

samples/interactivePersonalAccount.ts
sadasant: I'll use `authResponse`

samples/interactivePersonalAccount.ts
sadasant: Why I think this is important:
The sample can be misleading if users run it without changing `<my-subscription>`. The sample won't show what is expected.

samples/interactivePersonalAccount.ts
sadasant: It seems to me that the code is self explanatory regarding the fact that we are trying to get the subscriptions details. We also don't know how many network requests will fail besides the one we use in the sample.

Issue #103, opened 2020-09-21
Update readme for steps to create a credential if one already has a token


ramya-rao-a:
As seen in https://github.com/Azure/azure-sdk-for-js/issues/11243, when one already has an access token, there are no credential class in this package that one can use. This issue is to update the readme to include the workaround listed in the linked issue

daviwil:
Fixed in #104 and #105.

PR #104, opened 2020-09-24
Make buildTenantsList more resilient to error responses

### LINKED ISSUE #81: Cannot authenticate Azure SQL DB due to InvalidAuthenticationTokenAudience error (aud: `https://database.windows.net/`)
Hi,

we have a node app and we are trying to use Azure Active Directory Authentication for authentication with Azure SQL DB. we use [`sequelize`](https://sequelize.org/v5/class/lib/sequelize.js~Sequelize.html#instance-constructor-constructor) to establish the connection. under the hood, it uses [`tedious`](https://tediousjs.github.io/tedious/api-connection.html), which uses this package (`@azure/ms-rest-nodeauth`). Here are the versions we use:

- `nodejs` - v10.16.3
- `@azure/ms-rest-nodeauth` - v2.0.5
- `sequelize` - v5.18.4
- `tedious` - v6.4.0

The configuration we pass to `sequelize` is as follow:

```javascript
const Sequelize = require('sequelize');

const DB_CONFIG = {
    username: process.env.DB_USERNAME,
    password: process.env.DB_PASSWORD,
    name: process.env.DB_NAME,
    settings: {
        host: process.env.DB_HOST,
        port: process.env.DB_PORT,
        dialect: 'mssql',
        dialectOptions: {
            authentication: {
              type: 'azure-active-directory-password',
            },
            options: {
                encrypt: true,
            },
            requestTimeout: 600000,
        },
        pool: {
            max: 10,
            min: 0,
            idle: 5000,
            acquire: 5000,
        },
    },
};

const sequelize = new Sequelize(
    DB_CONFIG.name,
    DB_CONFIG.username,
    DB_CONFIG.password,
    DB_CONFIG.settings
);
```

however, we are facing an authentication problem coming from `@azure/ms-rest-nodeauth`. The error that gets thrown seems some sort of improper handling of the response from `https://management.azure.com/tenants?api-version=2016-06-01`:
```
TypeError: tenants.value is not iterable
```
from the [for loop](https://github.com/Azure/ms-rest-nodeauth/blob/adea8415de97a8076006fbd9948123d09dfdd78d/lib/subscriptionManagement/subscriptionUtils.ts#L94) inside `buildTenantList` method (inside `subscriptionUtils.ts`), which expects `value` to be always present and iterable. Nevertheless, the actual problem is the response from the above call - the audience doesn't match. The actual response body (`res.parsedBody`) is:

```json
{
  error: {
    code: 'InvalidAuthenticationTokenAudience',
    message: 'The access token has been obtained for wrong audience or resource \'https://database.windows.net/\'. It should exactly match with one of the allowed audiences \'https://management.core.windows.net/\',\'https://management.core.windows.net\',\'https://management.azure.com/\',\'https://management.azure.com\'.',
  },
}
``` 

The audience in the `accessToken` we get is `https://database.windows.net/` (which makes sense to me since we are trying to authenticate against a DB, but might be wrong?). We extracted the token from the [`creds.getToken`](https://github.com/Azure/ms-rest-nodeauth/blob/adea8415de97a8076006fbd9948123d09dfdd78d/lib/login.ts#L176) call inside `withUsernamePasswordWithAuthResponse`. The response from `creds.getToken` also contains our tenant id (`tenantId`). However, the [next line](https://github.com/Azure/ms-rest-nodeauth/blob/adea8415de97a8076006fbd9948123d09dfdd78d/lib/login.ts#L179) (`const tenantList = await buildTenantList(creds);`) is the one that fails with the above mentioned errors. If we try to manually modify these two lines (inside `withUsernamePasswordWithAuthResponse` and grab the tenant id and assign it to `tenantList`, we have no problem connecting to the DB and executing queries from our node app:
```javascript
// ...
  const tokenResponse = await creds.getToken();

  const tenantList = [ tokenResponse.tenantId ];
// ...
```

so our question(s) are:
1. is there a need to build the tenant list in all situations or can that be grabbed from the response depending on the authorization request (e.g. DB auth) and skip a call to `https://management.azure.com/tenants?api-version=2016-06-01` or is this the intended behavior?
1. if the answer to the above is `we must always make that call and build the list`, is it possible to add `https://database.windows.net/` to the list of allowed audiences? if not, what can we do to successfully authenticate against our Azure SQL DB using Azure AD? (quite possible we are missing a step somewhere :))
1. is `https://database.windows.net/` the right audience? if not, is this something we can modify on our end in Azure (e.g. Azure AD setup, etc.)?

Thank you for looking into our inquiry!

daviwil:
This change fixes #81 which reports that `buildTenantsList` throws an error when attempting to list subscriptions.  The fix is to detect an error response and return an empty list instead.  I'm also fixing the `withUsernamePasswordWithAuthResponse` method so that we use the `tenantId` that gets returned by the token request when `buildTenantsList` returns an empty array.

### DIFF
--- a/lib/credentials/msiTokenCredentials.ts
+++ b/lib/credentials/msiTokenCredentials.ts
@@ -116,7 +116,7 @@ export abstract class MSITokenCredentials implements TokenClientCredentials {
       if (typeof parsedBody["expires_on"] === "string") {
         // possibly a Date string '09/14/2017 00:00:00 PM +00:00'
         if (parsedBody["expires_on"].includes(":") || parsedBody["expires_on"].includes("/")) {
-          parsedBody.expiresOn = new Date(parsedBody["expires_on"], 10);
+          parsedBody.expiresOn = new Date(parsedBody["expires_on"]);
         } else {
           // normal number as a string '1504130527'
           parsedBody.expiresOn = new Date(parseInt(parsedBody["expires_on"], 10));
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.0.6 - 2020/09/24
+
+- Fixed a bug where `buildTenantsList` will throw an error when it can't list tenants
+- Added instructions for authenticating with an existing token
+
 ## 2.0.5 - 2019/08/22
 - Fixed a bug where the callback to `loginWithServicePrincipalSecretWithAuthResponse` is sometimes not called.
 - Fix bug prevent tenant IDs from being discovered on auth
--- a/README.md
+++ b/README.md
@@ -200,6 +200,26 @@ function getUrl(subscriptionId: string): string {
 main();
 ```
 
+### Authenticating with an existing token
+
+If you have acquired a valid Azure Active Directory token from another source, you can use it to authenticate with Azure SDK libraries using the following code snippet:
+
+```js
+const { HttpHeaders } = require("@azure/ms-rest-js");
+
+function getCredentialForToken(accessToken) {
+  return {
+    signRequest: (request) => {
+      if (!request.headers) request.headers = new HttpHeaders();
+      request.headers.set("Authorization", `Bearer ${accessToken}`);
+      return Promise.resolve(request);
+    }
+  };
+}
+
+const creds = getCredentialForToken("your existing token");
+```
+
 ### Contributing
 
 This project welcomes contributions and suggestions.  Most contributions require you to agree to a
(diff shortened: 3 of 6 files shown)


daviwil:
Also including a fix for #103.  I'll cherry-pick these changes over to `master` once the PR is done.  Will investigate test failures in a bit.

lib/credentials/msiTokenCredentials.ts
```diff
@@ -116,7 +116,7 @@ export abstract class MSITokenCredentials implements TokenClientCredentials {
       if (typeof parsedBody["expires_on"] === "string") {
         // possibly a Date string '09/14/2017 00:00:00 PM +00:00'
         if (parsedBody["expires_on"].includes(":") || parsedBody["expires_on"].includes("/")) {
-          parsedBody.expiresOn = new Date(parsedBody["expires_on"], 10);
```
ramya-rao-a, line 119: What was this change about?

ramya-rao-a (approved):

PR #113, opened 2021-02-12
Simplify API reference docs by following TSDoc

ramya-rao-a:
The current format of  `@property`, `@param` and `@returns` tags where type information is added does not get rendered as needed

Example: 
image

Since the source code in this package uses TypeScript and not JavaScript, we can simplify the docs as follows
- Use `@returns` instead of `@return` as per TSDoc
- Avoid type information in `@property`, `@param` and `@returns` tags
- Remove the return tag if there is no info to provide other than type info
- Remove tags like `@constructor`, `memberof`,  `@interface` and `@class` that are not needed in TS files

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 3.0.7 - 2021/02/23
-- Updated doc comments on all exported members to follow TSDoc for better API reference documentations.
+- Updated doc comments on all exported members to follow TSDoc for better API reference documentation.
 
 ## 3.0.6 - 2020/09/25
 - Fixed a bug where `buildTenantsList` will throw an error when it can't list tenants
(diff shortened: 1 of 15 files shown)


sadasant (approved):
This is way better! Thank you 🙌

joheredi (approved):

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.0.7 - 2021/02/16
```
ramya-rao-a, line 3:
```suggestion
## 3.0.7 - 2021/02/23
```

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.0.7 - 2021/02/23
+- Updated doc comments on all exported members to follow TSDoc for better API reference documentations.
```
ramya-rao-a, line 4:
```suggestion
- Updated doc comments on all exported members to follow TSDoc for better API reference documentation.
```

PR #120, opened 2021-03-26
Add prettier and format code

ramya-rao-a:
Adding prettier and formatting code just like https://github.com/Azure/ms-rest-js/pull/415


sadasant (approved):
Looks _prettier_!

PR #121, opened 2021-03-26
Stengthen types around AuthResponse

ramya-rao-a:
This PR strengthens the types around AuthReponse inspired by the changes in #75
Creating this PR as we did not get a response to https://github.com/Azure/ms-rest-nodeauth/pull/75#issuecomment-770492773

Additionally, this updates the `execAz` command to return a rejected promise when there is no output from running the az command

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.1.0 - Unreleased
+
+- Use stronger types in places where AuthResponse was used in return values that used the abstract base class `TokenCredentialsBase`. See [PR 121](https://github.com/Azure/ms-rest-nodeauth/pull/121) for details.
+
 ## 3.0.10 - 2021/05/20
 
 - Updated the dependency `adal-node` to version `^0.2.0`. This fixes customer issue: [125](https://github.com/Azure/ms-rest-nodeauth/issues/125).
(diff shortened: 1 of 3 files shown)


sadasant (approved):
Looks good!

arthurschreiber:
@ramya-rao-a Ah, sorry - I totally missed your message over in #75!

You might have to release these changes as semver-major, if the version numbers of this package are following semantic versioning. This is because I think some of these type changes are breaking backwards compatibility (but they're more correct than before).

Specifically, the change to the return type of `withAuthFile()` does not seem to be backwards compatible.

ramya-rao-a:
Hey @arthurschreiber 

Can you elaborate on why the change in the return type for `withAuthFile()` would be breaking while similar changes to the return types for other method would not be?

Previously, the return type would be around `TokenCredentialsBase` and now it would be `ApplicationTokenCredentials | ApplicationTokenCertificateCredentials`. Both `ApplicationTokenCredentials` and `ApplicationTokenCertificateCredentials` extend `TokenCredentialsBase`.

If we were changing the return type from say `ApplicationTokenCredentials` to `ApplicationTokenCredentials | ApplicationTokenCertificateCredentials`, then I can see how the change would be breaking because not all properties of `ApplicationTokenCredentials` are available in `ApplicationTokenCertificateCredentials`

@sadasant, I see you have upvoted Arthur's concerns above. Do you have any insights on how the changes to the `withAuthFile` is breaking?

sadasant:
@ramya-rao-a no, I think you’re correct, I just wanted to validate the feedback somehow. Feedback is always good!

PR #123, opened 2021-03-30
Arguments passed to execFile should be array of strings

### LINKED ISSUE #122: Breaking Change for AzureCliCredentials
**Describe the bug**
I have just run a npm install using this package to find a nasty surprise that i cannot log in with `AzureCliCredentials.create({resource})` for mac and linux (not tested windows)_

Looking at the source code i have found that a bug  in the following commit 
https://github.com/Azure/ms-rest-nodeauth/commit/1b8dcfb6d7e1c8a0a00044286021a841961cf636

```
 exec(`az ${cmd} --out json`, { encoding: "utf8" }, (error, stdout) => {
to
    execFile(`az`, [cmd, `--out json`], { encoding: "utf8" }, (error, stdout) => {
```

the issue being that execFile args with spaces does not work. Which i have also been able to reproduce from terminal 
```
az "account show"
'account show' is misspelled or not recognized by the system.
Did you mean 'account' ?

``` 

However if i correct execFile without spaces, it works, not very nice looking
```
execFile(`az`, ['account', 'show', '--out', 'json']
```

**To Reproduce**

await AzureCliCredentials.create({resource});

```
Error: Command failed: az account show --out json
ERROR: 'account show' is misspelled or not recognized by the system.
Did you mean 'account' ?

TRY THIS:
https://aka.ms/cli_ref
Read more about the command in reference docs

```

**Expected behavior**
It should execute the commands without error, as this impacted all 

**work around**

I have downgraded my package

ramya-rao-a:
Fixes #122 by updating the execFile method to take array of strings representing the cmd arguments

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,19 +1,25 @@
 # Changelog
 
-## 3.0.9 - Unreleased
+## 3.0.9 - 2021/03/30
+
 - Pass user provided `tokenCache` option to `withUsernamePasswordWithAuthResponse` and `withServicePrincipalSecretWithAuthResponse` methods to the credentials being created.
+- Fix issue with `AzureCliCredentials` where the `az` command fails due to spaces in the argument list passed to `execFile`. This regression got introduced in the previous update.
 
 ## 3.0.8 - 2021/03/23
+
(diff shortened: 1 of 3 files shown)


xirzec (approved):
Looks great! I'm glad you figured out the problem on Windows!

Issue #127, opened 2020-09-30
Ref doc issue - ms-rest-nodeauth loginWithAppServiceMSI()


diberry:
This [ref doc](https://docs.microsoft.com/en-us/javascript/api/%40azure/ms-rest-nodeauth/?view=azure-node-latest#loginwithappservicemsi--) method doesn't have enough information to use. 
* Is there supposed to be a `%40` before azure in the URL? 
* Is there more information about the AppServiceMSI or a link to some other doc? 
* 3rd paragraph looks incorrectly formatted

![image](https://user-images.githubusercontent.com/41597107/94735753-11f46c00-0320-11eb-9b07-e735676699a2.png)

ramya-rao-a:
> Is there supposed to be a %40 before azure in the URL?

Not required. Do you recall which link you clicked to get here? That can tell us what needs to be fixed where

> Is there more information about the AppServiceMSI or a link to some other doc?

If you have an aka.ms link that is language agnostic, that would be useful

> 3rd paragraph looks incorrectly formatted

Created #133

ramya-rao-a:
@sadasant Please close this issue after ensuring the docs have been updated post releasing 3.1.0 of `@azure/ms-rest-nodeauth`

sadasant:
Released! https://www.npmjs.com/package/@azure/ms-rest-nodeauth/v/3.1.0

sadasant:
The docs haven’t been updated. I’m investigating.

sadasant:
Ok now it’s fixed! https://docs.microsoft.com/en-us/javascript/api/%40azure/ms-rest-nodeauth/?view=azure-node-latest#loginWithAppServiceMSI__ closing

PR #133, opened 2021-07-21
Fix docs for return types in exported APIs

ramya-rao-a:
This PR fixes the docs issue mentioned in #127

This should have been done as part of #113 but was missed.


sadasant (approved):
👍

PR #143, opened 2022-02-11
Migration to Identity: Issues in the code snippets

sadasant:
This PR fixes two issues in the code snippets:

- We passed the `tokenAudience` to the `ClientSecretCredential` as a parameter when `ClientSecretCredential` does not receive such parameter.
- We were not calling the `getToken` method of the credential used in the sample for the `Regarding scopes` section. We were instead passing it to a credential.
- In the `Regarding scopes` code snippet, we were also moving from `interactiveLogin` to `AzureCliCredential`, which is confusing. Using `InteractiveBrowserCredential` instead.

### DIFF
--- a/migrate-to-identity-v2.md
+++ b/migrate-to-identity-v2.md
@@ -108,7 +108,7 @@ const credential = new ClientSecretCredential(
 You will continue specifying a `baseUri` when creating the client in the Azure package to point to the correct scope relative to a national cloud. An example follows:
 
 ```diff
-- import { ApplicationTokenCredentials } from "@azure/ms-rest-nodeauth";
+- import { loginWithServicePrincipalSecret } from "@azure/ms-rest-nodeauth";
 + import { ClientSecretCredential, AzureAuthorityHosts } from "@azure/identity";
 - import { Environment } from "@azure/ms-rest-azure-env";
 import { SubscriptionClient } from "@azure/arm-subscriptions";
@@ -119,16 +119,14 @@ import * as dotenv from "dotenv";
 dotenv.config();
 
 const clientId = process.env.AZURE_CLIENT_ID;
-const domain = process.env.AZURE_TENANT_ID;
-const tenantId = process.env.AZURE_TENANT_ID;
-const secret = "process.env.AZURE_CLIENT_SECRET;
-- const tokenAudience = "https://management.chinacloudapi.cn/";
+const domain = process.env.AZURE_TENANT_ID; // domain or tenantId
+const secret = process.env.AZURE_CLIENT_SECRET;
 - const environment = Environment.ChinaCloud;
++ const authorityHost = AzureAuthorityHosts.AzureChina;
+
 async function main() {
-- const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
-+ const credential = new ClientSecretCredential(tenantId, clientId, secret, {
-+  authorityHost: AzureAuthorityHosts.AzureChina
-+ });
+- const credential = await loginWithServicePrincipalSecret(clientId, secret, domain, { environment });
++ const credential = new ClientSecretCredential(domain, clientId, secret, { authorityHost });
   const client = new SubscriptionClient(credential, {
 -  baseUri: environment.resourceManagerEndpointUrl,
 +  baseUri: "https://management.chinacloudapi.cn"
--- a/migrate-to-identity-v2.md
+++ b/migrate-to-identity-v2.md
@@ -127,7 +127,7 @@ const tokenAudience = "https://graph.microsoft.com/";
 
 async function main() {
 - const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
-+ const credential = new ClientSecretCredential(tenantId, clientId, secret, tokenAudience, {
++ const credential = new ClientSecretCredential(tenantId, clientId, secret, {
 +  authorityHost: AzureAuthorityHosts.AzureChina
 + });
   const client = new SubscriptionClient(credential, {
@@ -221,7 +221,8 @@ async function main() {
 -  const authResponse = await interactiveLogin({
 -    tokenAudience: "https://vault.azure.net/"
 -  });
-+  const credential = new AzureCliCredential("https://vault.azure.net/.default");
++  const credential = new InteractiveBrowserCredential();
++  const accessToken = await credential.getToken("https://vault.azure.net/.default");
 }
 
 main().catch(console.error);
--- a/migrate-to-identity-v2.md
+++ b/migrate-to-identity-v2.md
@@ -110,7 +110,7 @@ You will continue specifying a `baseUri` when creating the client in the Azure p
 ```diff
 - import { loginWithServicePrincipalSecret } from "@azure/ms-rest-nodeauth";
 + import { ClientSecretCredential, AzureAuthorityHosts } from "@azure/identity";
-- import { Environment } from "@azure/ms-rest-azure-env";
+import { Environment } from "@azure/ms-rest-azure-env";
 import { SubscriptionClient } from "@azure/arm-subscriptions";
 
 import * as msRest from "@azure/ms-rest-js";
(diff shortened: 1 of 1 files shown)


migrate-to-identity-v2.md
```diff
@@ -127,7 +127,7 @@ const tokenAudience = "https://graph.microsoft.com/";
 
 async function main() {
 - const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
```
ramya-rao-a: The token audience is set for graph. Why are we passing that when creating the credential from ms-rest-nodeauth in the first place if the sample is about creating a credential to be passed to SubscriptionClient?

migrate-to-identity-v2.md
```diff
@@ -221,7 +221,8 @@ async function main() {
 -  const authResponse = await interactiveLogin({
 -    tokenAudience: "https://vault.azure.net/"
 -  });
-+  const credential = new AzureCliCredential("https://vault.azure.net/.default");
++  const credential = new InteractiveBrowserCredential();
++  const accessToken = await credential.getToken("https://vault.azure.net/.default");
```
ramya-rao-a, line 222: If we are showing how to get the access token using azure/identity, shouldnt we also show the same for ms-rest-nodeauth here?

sadasant:
I’ll follow up tomorrow!

migrate-to-identity-v2.md
sadasant: how does it look now? I’m using the china cloud as the audience.

migrate-to-identity-v2.md
sadasant: How does it look now? interactiveLogin returns the device code credential (in plural `credentials` is more commonly called in `ms-rest-nodeauth`). That credential then has a similar getToken method, but one that does not receive parameters.

migrate-to-identity-v2.md
ramya-rao-a: Can we test it out?

sadasant: sure can! (tomorrow, but will do)

sadasant: This works for me if I use the AzureCloud instead of the China Cloud, but it works 🤔

Here’s my code:

```ts
const tokenAudience = "https://management.azure.com/";
const environment = Environment.AzureCloud;
const clientId = process.env.AZURE_CLIENT_ID;
const domain = process.env.AZURE_TENANT_ID;
const secret = process.env.AZURE_CLIENT_SECRET;
const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
console.log(await credential.getToken());
```

ramya-rao-a: This code snippet we are editing is about getting subscriptions from the SubscriptionClient while your code snippet above gets the token...

sadasant: This works:

```ts
const tokenAudience = "https://management.azure.com/";
const environment = Environment.AzureCloud;
const clientId = process.env.AZURE_CLIENT_ID;
const domain = process.env.AZURE_TENANT_ID;
const secret = process.env.AZURE_CLIENT_SECRET;
const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
const client = new SubscriptionClient(credential, {
  baseUri: environment.resourceManagerEndpointUrl,
  baseUri: "https://management.azure.com/"
});
const subscriptions = await client.subscriptions.list();
```

Which I was expecting, since I was able to get a token in the previous snippet I tried 🤔 Is there something specific that I’m missing?

ramya-rao-a: I am still slightly confused...

Take the below facts:
- `tokenAudience` is used to determine which "resource" to use and is an optional argument
- This happens in [getActiveDirectoryResourceId()](https://github.com/Azure/ms-rest-nodeauth/blob/3710a14c26a584dd667a0f726fe015f868d8cdcd/lib/credentials/tokenCredentialsBase.ts#L59) where you can see that the default value that gets used is Environment.activeDirectoryResourceId which for public cloud is `https://management.core.windows.net/` and for china cloud is `https://management.core.chinacloudapi.cn/`
- So, am confused why things are working when you pass `https://management.azure.com/` as the tokenAudience which is different from the default value that works for all mgmt plane packages.
- Also, from my understanding, a user only ever had to pass tokenAudience if they were using a data plane library

What I am trying to get at is whether we need to show the use of tokenAudience in this code snippet at all.

I then realized that you have to do that because for the China Cloud example, you need to pass the environment, but tokenAudience is the optional parameter that appears in the constructor before environment.

Which then makes me wonder if we should use the `withServicePrincipalSecret()` method here which uses an options bag instead of `new ApplciationTokenCredentials`

sadasant: Ohh I see now! I’ll take your advice, I will use `withServicePrincipalSecret()`.

sadasant: Alright, updated! How does it look now?

migrate-to-identity-v2.md
```diff
@@ -119,17 +119,14 @@ import * as dotenv from "dotenv";
...
 async function main() {
-- const credential = new ApplicationTokenCredentials(clientId, domain, secret, tokenAudience, environment);
-+ const credential = new ClientSecretCredential(tenantId, clientId, secret, tokenAudience, {
-+  authorityHost: AzureAuthorityHosts.AzureChina
-+ });
+- const credential = await loginWithServicePrincipalSecret(clientId, secret, domain, { environment });
++ const credential = new ClientSecretCredential(domain, clientId, secret, { authorityHost });
   const client = new SubscriptionClient(credential, {
 -  baseUri: environment.resourceManagerEndpointUrl,
 +  baseUri: "https://management.chinacloudapi.cn"
```
ramya-rao-a: Do we need to hard code this? Can we not fetch the value from `environment`?

sadasant: I’ll do it! — just feels kinda weird. We don’t generally recommend using ms-rest-azure-env with the new Identity package. On my way!

sadasant: Done!

migrate-to-identity-v2.md
```diff
@@ -108,9 +108,9 @@ const credential = new ClientSecretCredential(
 You will continue specifying a `baseUri` when creating the client in the Azure package to point to the correct scope relative to a national cloud. An example follows:
 
 ```diff
-- import { ApplicationTokenCredentials } from "@azure/ms-rest-nodeauth";
+- import { loginWithServicePrincipalSecret } from "@azure/ms-rest-nodeauth";
 + import { ClientSecretCredential, AzureAuthorityHosts } from "@azure/identity";
-- import { Environment } from "@azure/ms-rest-azure-env";
+import { Environment } from "@azure/ms-rest-azure-env";
```
ramya-rao-a, line 113: `Environment` is not needed when using `@azure/identity`
```suggestion
- import { Environment } from "@azure/ms-rest-azure-env";
```

migrate-to-identity-v2.md
```diff
@@ -119,20 +119,16 @@ import * as dotenv from "dotenv";
...
 
 const clientId = process.env.AZURE_CLIENT_ID;
-const domain = process.env.AZURE_TENANT_ID;
-const tenantId = process.env.AZURE_TENANT_ID;
-const secret = "process.env.AZURE_CLIENT_SECRET;
-const tokenAudience = "https://graph.microsoft.com/";
-- const environment = Environment.ChinaCloud;
+const domain = process.env.AZURE_TENANT_ID; // domain or tenantId
+const secret = process.env.AZURE_CLIENT_SECRET;
+const environment = Environment.ChinaCloud;
```
ramya-rao-a, line 124: `environment` is not needed when using `@azure/identity`
```suggestion
- const environment = Environment.ChinaCloud;
```

migrate-to-identity-v2.md
sadasant: I’m using it to set the `baseUri: environment.resourceManagerEndpointUrl,`, in order to not have the raw value.

migrate-to-identity-v2.md
sadasant: I’m using it to set the `baseUri: environment.resourceManagerEndpointUrl,`, in order to not have the raw value.

migrate-to-identity-v2.md
ramya-rao-a: Hmm... that might be problematic because eventually we want to deprecate all the `@azure/ms-rest*` packages. But thats a problem for another day :)

ramya-rao-a (approved):

migrate-to-identity-v2.md
ramya-rao-a: That is true, I didnt realize that when I recommended using `environment`.

I would be ok if we want to go back to hard-coding, but this time pull the value of `resourceManagerEndpointUrl` out into a constant

sadasant: For migration purposes, we don’t have a way for people to get this baseUri with the new libraries. They generally wouldn’t need to do this. I’m wondering if we should keep ms-rest-env as a useful tool only within this migration guide.

sadasant: I think I should merge this for now. We can revisit this later.

